PR #100, opened 2013-07-11
Ldap user attribute support

asheepapart:
This will allow slightly more customization of how principals are referenced in ldap; specifically, if your primary rdn is not by uid, but rather by full name (e.g. uid=bvanevery versus cn=Benjamin VanEvery).  

Overall this code should be restructured to use a bind user to create the initial context and then use that context to look up the user. The returned user object can then be used to determine the relative distinguished name, which can be combined with the search base and then combined with the provided user password to create a new context. This described method would be more flexible for fitting on top of other types of ldap schemas.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,5 @@ Gemfile.lock
 .idea_modules
 
 /conf/solr/data/
-/conf/dev_authentication.conf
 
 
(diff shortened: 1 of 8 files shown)


maddalab:
@asheepapart I took you description in the second para and implemented it [1]. I am going to close this PR looking forward to address any of your concerns in [1]

[1] https://github.com/tumblr/collins/pull/291

asheepapart:
Great, thanks!

PR #204, opened 2014-08-26
Bump httparty and restrict rspec to version 2

wwkeyboard:
I'm not sure if there is another reason to restrict the version of httparty, but it's causing conflicts with the sift gem(and I assume anything using a more modern version of httparty). I also restricted rspec because they have bumped major versions and made breaking changes to the api.


maddalab:
@defect is this still an issue with you PR in place now?

defect:
@maddalab That was just for collins_state. I'm planning on doing the same cleanup for collins_client. I'll make sure it works with the sift gem then.

PR #246, opened 2015-01-07
Will collins notify slack

Primer42:
Adding slack support to the `collins-notify` gem

@byxorna @maddalab @CaptPhunkosis


byxorna:
@Primer42 can you rebase and squash your commits?

Primer42:
yep.

PR #247, opened 2015-01-07
Will collins notify slack

Primer42:
Squashed version of #246 
Adding slack support to `collins-notify`
@byxorna @maddalab @CaptPhunkosis


support/ruby/collins-notify/Gemfile
```diff
@@ -7,10 +7,11 @@ gem 'mail',           '~> 2.4.4'
...
-  gem "rdoc", "~> 3.12"
-  gem "bundler", "~> 1.2.0"
-  gem "jeweler", "~> 1.8.4"
-  gem "simplecov"
+  gem 'i18n',         '< 0.7.0'	# versions above 0.7.0 require ruby 1.9.3
+  gem 'rspec',        '~> 2.12.0'
+  gem 'yard',         '~> 0.8.3'
+  gem 'rdoc',         '~> 3.12'
+  gem 'bundler',      '>= 1.2.0'
+  gem 'jeweler',      '~> 1.8.4'
```
byxorna: :( can we use this opportunity to remove this?

Primer42: Yes

support/ruby/collins-notify/lib/collins_notify/adapter/slack.rb
```diff
@@ -0,0 +1,85 @@
...
+          slack_hash[op] = config.adapters[:slack][op]
+        end
+      end
+
+      slack_hash['text'] = get_message_body(binding)
+
+      @logger.debug "slack parameters: #{slack_hash.inspect}"
+        
+      if config.test? then
+        @logger.info "Not sending message in test mode"
```
byxorna: Log what message you arent sending, instead of the slack parameters debug above?

support/ruby/collins-notify/lib/collins_notify/adapter/slack.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    # slack can both DM users and post to channels
+    # If the recipient isn't explicitly a user (@username) then we'll assume the string is a channel
+    def make_channel chan
+      @logger.debug "In make_channel, got #{chan}"
+      # see if we're DMing a user, or posting to a channel
+      chan_type_identifier = chan.start_with?('@') ? '@' : '#'
+      # ensure the channel name starts with the right character
+      chan = chan.start_with?(chan_type_identifier) ? chan : "##{chan}"
```
byxorna: I think this whole function logic could just be replaced with ```chan.gsub(/^(?!#|@)/,'#')```

```
irb(main):007:0> "nohash".gsub(/^(?!#|@)/,'#')
=> "#nohash"
irb(main):008:0> "@user".gsub(/^(?!#|@)/,'#')
=> "@user"
irb(main):009:0> "#channel".gsub(/^(?!#|@)/,'#')
=> "#channel"
```

byxorna:
Aside from those small points, looks great!

support/ruby/collins-notify/lib/collins_notify/adapter/slack.rb
Primer42: Yep you're right.

support/ruby/collins-notify/lib/collins_notify/adapter/slack.rb
Primer42: I'm not so sure.... if you aren't asking for verbose output, I don't want to print a more information than necessary, and all the other adapters user more or less this string as an info message if you're in test mode.

support/ruby/collins-notify/lib/collins_notify/adapter/slack.rb
```diff
@@ -0,0 +1,84 @@
...
+      end
+    end
+
+    # slack can both DM users and post to channels
+    # If the recipient isn't explicitly a user (@username) then we'll assume the string is a channel
+    def make_channel chan
+      @logger.debug "In make_channel, got #{chan}"
+      # If we are DMing a user (chan looks like @username), do nothing
+      # otherwise, make sure chan starts with a '#', if it doesn't already
+      chan = chan.gsub(/^(?!#|@)/,'#')
```
roymarantz: Why #gsub instead of #sub since you are are using a ^ anchor I don't think it can match multiples anyway???

support/ruby/collins-notify/Gemfile
```diff
@@ -7,10 +7,10 @@ gem 'mail',           '~> 2.4.4'
 gem 'nokogiri',       '~> 1.5.2'
 
 group :development do
-  gem "rspec", "~> 2.12.0"
-  gem "yard", "~> 0.8.3"
-  gem "rdoc", "~> 3.12"
-  gem "bundler", "~> 1.2.0"
-  gem "jeweler", "~> 1.8.4"
-  gem "simplecov"
+  gem 'i18n',         '< 0.7.0'	# versions above 0.7.0 require ruby 1.9.3
```
roymarantz: Do you want to require ruby < 1.9.3 too in this Gemfile?
You can also test the version of ruby and adjust the version based on that.

Primer42: We used to do that for other gems, but that was causing problems for people.  I don't think we should do that here.
https://github.com/tumblr/collins/pull/150

support/ruby/collins-notify/lib/collins_notify/adapter/slack.rb
```diff
@@ -0,0 +1,84 @@
...
+      end
+    end
+
+    # slack can both DM users and post to channels
+    # If the recipient isn't explicitly a user (@username) then we'll assume the string is a channel
+    def make_channel chan
+      @logger.debug "In make_channel, got #{chan}"
+      # If we are DMing a user (chan looks like @username), do nothing
+      # otherwise, make sure chan starts with a '#', if it doesn't already
+      chan = chan.sub(/^(?!#|@)/,'#')
```
byxorna: no need for this temp variable. All you need is:
```
def make_channel chan
  @logger.debug "In make_channel, got #{chan}"
  chan.sub(/^(?!#|@)/,'#')
end
```

byxorna:
Awesome, lgtm!

support/ruby/collins-notify/lib/collins_notify/adapter/slack.rb
Primer42: I just found specs for valid characters in channel names - I'm adding a commit no enforce those now

PR #249, opened 2015-01-11
allow us to do trusted builds for the container

byxorna:
Lets try this again.

@Primer42 @maddalab @roymarantz  
This changes the container over to using the supported java:7 container, which includes the JCE policy files. The build is much simpler now, and allows us to setup the collins container as a trusted build (read: automatic) instead of manual as it is now.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.git
+_site
+target
+support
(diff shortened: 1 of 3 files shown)


Primer42:
LGTM :snowman: :clock12: :confetti_ball:

PR #253, opened 2015-01-21
Add Collins::Asset::count() method for counting assets based on selector

KirilAngov:
Adds an optimized alternative to `Collins::Asset::find()` which only returns the total number of assets found.

@byxorna


support/ruby/collins-client/lib/collins/api/asset.rb
```diff
@@ -123,6 +123,40 @@ def find_similar asset_or_tag, size = 50, sort = "ASC", sort_type = "distance",
...
+    # Count number of assets matching the specified criteria
+    #
+    # @param [Hash] options Query options (same as in the "find" method)
+    # @return integer The number of assets matching the query
+    # @raise [UnexpectedResponseError] If the HTTP response code is not a 200
+    def count options = {}
+      # create a copy so that we do not modify the original options array
+      options = options.dup
+
+      if options.include? "size" or options.include? "page"
```
byxorna: does options hash automatically convert symbol keys to strings for this ```#include?``` operator?

support/ruby/collins-client/lib/collins/api/asset.rb
```diff
@@ -123,6 +123,40 @@ def find_similar asset_or_tag, size = 50, sort = "ASC", sort_type = "distance",
...
+      else
+        options["size"] = 1
+        options["page"] = 0
+      end
+
+      query = asset_hash_to_find_query options
+      params = query.to_a.map do |param|
+        key, val = param
+        if val.is_a?(Array)
+          val.map{|v| "#{key}=#{asset_escape_attribute(v)}"}.join("&")
```
byxorna: Seems like this kind of logic should be encapsulated elsewhere, perhaps in a util module?

support/ruby/collins-client/lib/collins/api/asset.rb
KirilAngov: :+1: 

```
irb(main):001:0> options = {page: 1, size: 10}
=> {:page=>1, :size=>10}
irb(main):002:0> options.include? "page"
=> false
irb(main):003:0> options.include? :page
=> true
```

byxorna: Magic is the worst.

KirilAngov:
Can I get some thumbs up?

maddalab:
lgtm

byxorna:
@KirilAngov any thoughts on pulling the query params string assembly out of your method into a helper function somewhere?

support/ruby/collins-client/lib/collins/api/asset.rb
```diff
@@ -123,6 +123,40 @@ def find_similar asset_or_tag, size = 50, sort = "ASC", sort_type = "distance",
...
+
+      if options.include? :size or options.include? :page
+        raise ExpectationFailedError.new "Do not specify 'size' or 'page' options when counting assets"
+      else
+        options[:size] = 1
+        options[:page] = 0
+      end
+
+      query = asset_hash_to_find_query options
+      params = query.to_a.map do |param|
```
roymarantz: any reason not to use map on the hash directly?  i.e. query.map do |key, val|

support/ruby/collins-client/lib/collins/api/asset.rb
KirilAngov: You both have good points but this is 1:1 almost from the find method. I am willing to improve them both in a different pull request if you think it is needed.

byxorna: Ok, fair point. Separate PR is better. :sun_with_face:

byxorna:
LGTM!

KirilAngov:
merge anyone?

roymarantz:
You've already gotten a couple of thumbs up.  Here's another :+1: 
I does look like the Travis CI needs some work.

byxorna:
Yea Travis won't go Green cause the Scala tests are broken. I'll merge

PR #259, opened 2015-01-29
DAEMON is no longer used, stop asking for it during setup.

pdex:


byxorna:
Awesome, thanks @pdex! This slipped through the cracks when we were ripping out daemon.

Primer42:
Sweet! LGTM :snowman: :clock12: :confetti_ball:

PR #261, opened 2015-01-30
add status:state label on asset show pages

byxorna:
@maddalab @Primer42 This makes it easier to see what status an asset is in, even if you are on another menu pill. Nice to have to prevent someone from rebooting the wrong box when clicking through tabs :)
The label has a hover of the status+state description, just to be a fancy pants. It is also right justified to the top+right of the page header, so it looks clean (not terribly obvious from the screen shots).

screen shot 2015-01-30 at 3 58 16 pm
screen shot 2015-01-30 at 3 58 43 pm
screen shot 2015-01-30 at 3 59 18 pm
screen shot 2015-01-30 at 3 59 53 pm
screen shot 2015-01-30 at 4 02 22 pm

### DIFF
--- a/app/util/views/Summary.scala
+++ b/app/util/views/Summary.scala
@@ -2,6 +2,7 @@ package util
 package views
 
 import models.Asset
+import models.Status
 
 import play.api.mvc.Content
 import play.api.templates.Html
@@ -29,7 +30,8 @@ trait Summarizer[T] {
 case class AssetSummary(template: String) extends Summarizer[Asset] {
   override val replacementMap = Map(
     "assetType" -> getAssetType _,
-    "hostname" -> getHostname _
+    "hostname" -> getHostname _,
+    "statusLabel" -> getStatusLabel _
   )
 
   def getAssetType(asset: Asset): String = {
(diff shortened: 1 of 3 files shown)


Primer42:
LGTM :snowman: :clock12: :confetti_ball:

maddalab:
lgtm

app/util/views/Summary.scala
```diff
@@ -38,6 +40,27 @@ case class AssetSummary(template: String) extends Summarizer[Asset] {
...
+    }
+    val badge_class = asset.getStatus.name match {
+      case "Allocated"                    => "label-primary"
+      case "Cancelled" | "Decommissioned" => "label-default"
+      case "Maintenance"                  => "label-danger"
+      case "Incomplete" | "New"           => "label-warning"
+      case "Provisioned" | "Provisioning" => "label-info"
+      case "Unallocated"                  => "label-success"
+      case _                              => "label-default"
+    }
```
roymarantz: too bad this is in code instead of a config file :-(

byxorna: Agreed. It would be way more hassle than its worth to break colors out into config files. I dont like it either, but... I didnt want to spend all day doing this. :)

byxorna:
Any objections to merging in? Anyone want me to put up a container that has this change in it to try it out?

app/util/views/Summary.scala
Primer42: Well, these are CSS tags right? So at the very least someone could customize using custom CSS.

Primer42:
I built this from source (ya, I'm just old :school: like that) and it worked fine.  Merge away!

byxorna:
Sweet. :boom: :fire: :rice_cracker: :zap: :rotating_light:

PR #263, opened 2015-02-04
replaced Twitter futures with Scala futures

nsauro:
removed all Awaits except for one that is used by
MultiCollins/BackgroundProcess

This is one PR which removes Twitter futures and uses Scala Futures/timeouts.  I removed all timeouts that were not connected to multi collins(not used), and for background processing(because changing BackgroundProcess to being asynchronous is not a trivial task).

I have another branch off of this one which removes the finagle library entirely.  I will pr that one separately as I wanted to keep this one focused.

### DIFF
--- a/app/collins/softlayer/SoftLayerPlugin.scala
+++ b/app/collins/softlayer/SoftLayerPlugin.scala
@@ -10,11 +10,13 @@ import play.api.libs.json._
 import com.twitter.finagle.Service
 import com.twitter.finagle.builder.ClientBuilder
 import com.twitter.finagle.http.{Http, RequestBuilder, Response}
-import com.twitter.util.Future
+import com.twitter.util.{Future => TFuture, Throw => TThrow, Return => TReturn}
+import scala.concurrent.{ExecutionContext, Future, Promise}
 import org.jboss.netty.buffer.ChannelBuffers
 import org.jboss.netty.handler.codec.http.{HttpRequest, HttpResponse, QueryStringEncoder}
 import org.jboss.netty.util.CharsetUtil.UTF_8
 import scala.util.control.Exception.allCatch
+import play.api.libs.concurrent.Execution.Implicits._
 
 class SoftLayerPlugin(app: Application) extends Plugin with SoftLayer {
   type ClientSpec = ClientBuilder.Complete[HttpRequest, HttpResponse]
@@ -76,7 +78,7 @@ class SoftLayerPlugin(app: Application) extends Plugin with SoftLayer {
             case _ => 0L
           }
       }
-    } handle {
+    } recover  {
       case e => 0L
     }
   }
@@ -108,10 +110,10 @@ class SoftLayerPlugin(app: Application) extends Plugin with SoftLayer {
     doPowerOperation(e, "/SoftLayer_Hardware_Server/%d/rebootSoft.json")
   }
   override def verify(e: Asset): PowerStatus = {
-    Future.value(Failure("verify not implemented for softlayer"))
+    Future.successful(Failure("verify not implemented for softlayer"))
   }
   override def identify(e: Asset): PowerStatus = {
-    Future.value(Failure("identify not implemented for softlayer"))
+    Future.successful(Failure("identify not implemented for softlayer"))
   }
 
   override def activateServer(id: Long): Future[Boolean] = {
@@ -130,7 +132,7 @@ class SoftLayerPlugin(app: Application) extends Plugin with SoftLayer {
         case JsBoolean(v) => v
         case o => false
       }
-    } handle {
+    } recover {
       case e => false
     }
   }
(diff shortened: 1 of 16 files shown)


app/controllers/actors/AssetCancelProcessor.scala
```diff
@@ -2,43 +2,51 @@ package controllers
...
-                  Left(Api.getErrorMessage("There was an error cancelling this server"))
-                case ticketId =>
-                  Asset.inTransaction {
-                    MetaWrapper.createMeta(asset, Map("CANCEL_TICKET" -> ticketId.toString))
-                    AssetLifecycle.updateAssetStatus(asset, AStatus.Cancelled, None, reason)
-                  }
-                  Await.result(plugin.setNote(n, "Cancelled: %s".format(reason)))
-                  Right(ticketId)
-              }
+        SoftLayer.pluginEnabled.fold(softlayerDisabled){ plugin =>
```
nsauro: This change looks a lot bigger than it is.  I was having trouble following it, so I decided to move some of the constants into separate declarations.  I also opted to use the "fold" function instead of map/getOrElse.

app/util/concurrent/BackgroundProcessor.scala
```diff
@@ -36,19 +35,28 @@ object BackgroundProcessor {
...
+    val mpd: Future[RESPONSE] = f.map{x => result((None, Some(x)))}
+    mpd.recover {
+      case t : TimeoutException => result((Some(SexyTimeoutException(cmd.timeout)), None))
+      case th : Throwable => result((Some(th), None))
+    }
+  }
+
+  //possibly the WORST name ever.  Don't judge me.  'flatSend' because the above function is really just a fancy map function
+  //since you pass a func that goes from x->y and get a future[y], this is just the flatMap version of that.
+  def flatSend[PROC_RES,RESPONSE](cmd: BackgroundProcess[PROC_RES])(result: SendType[PROC_RES] => Future[RESPONSE])(implicit mf: Manifest[PROC_RES]): Future[RESPONSE] = {
```
nsauro: Yep...worst name ever.  I needed a send function which allowed me to flatMap on the result.

app/collins/softlayer/SoftLayerPlugin.scala
```diff
@@ -147,16 +149,27 @@ class SoftLayerPlugin(app: Application) extends Plugin with SoftLayer {
...
       true
-    } handle {
+    } recover {
       case e => false
     }
   }
 
   protected def makeRequest(request: HttpRequest): Future[HttpResponse] = {
     val client: Service[HttpRequest,HttpResponse] = clientSpec.build()
-    client(request) ensure {
```
maddalab: we should may be look into this in a different PR.

app/controllers/AssetWebApi.scala
```diff
@@ -9,6 +9,7 @@ import util.concurrent.BackgroundProcessor
 
 import play.api.mvc._
 import play.api.libs.json._
+import play.api.libs.concurrent.Execution.Implicits._
```
maddalab: Is this required?

app/controllers/actions/asset/ProvisionUtil.scala
```diff
@@ -246,7 +246,7 @@ trait Provisions extends ProvisionUtil with AssetAction { self: SecureAction =>
     import play.api.Play.current
 
     val ActionDataHolder(asset, pRequest, _, attribs) = adh
-    BackgroundProcessor.send(ProvisionerTest(pRequest)(request)) { res =>
+    BackgroundProcessor.send(ProvisionerTest(pRequest)(request, defaultContext)) { res =>
```
maddalab: I see the implicit imported, is the specification of defaultContext required?

app/controllers/actors/ProvisionerProcessor.scala
```diff
@@ -23,29 +22,33 @@ object ProvisionerStatus {
 
 case class ProvisionerResult(status: ProvisionerStatus, commandResult: CommandResult)
 
-case class ProvisionerTest(request: ProvisionerRequest, userTimeout: Option[FiniteDuration] = None)(implicit req: Request[AnyContent]) extends BackgroundProcess[ProvisionerResult]
+case class ProvisionerTest(request: ProvisionerRequest, userTimeout: Option[FiniteDuration] = None)(implicit req: Request[AnyContent], ec : ExecutionContext) extends BackgroundProcess[ProvisionerResult]
```
maddalab: i think making ec implicit will avoid the specification of defaultContext aboe

app/collins/softlayer/SoftLayerPlugin.scala
nsauro: The conversion of the the Twitter future to a Scala future?

app/controllers/AssetWebApi.scala
nsauro: AssetCancelProcessor needs an ExecutionContext since it is using Scala futures/Awaits to handle the Softlayer.cancelServer call.

app/controllers/actions/asset/ProvisionUtil.scala
nsauro: Yes, only because the second parameter list for implicits is being manually invoked because 'request' is not an implicit value.

app/collins/softlayer/SoftLayerPlugin.scala
nsauro: If its any help, I have another PR, right behind this which changes this class to not use the finagle library, and only to use the Play WS library.  That scraps all this conversion code, as the WS library returns Scala futures natively.

byxorna:
This doesnt have any test coverage, right? I believe not, as @maddalab showed with his 2.1.5 PR. Do you think its worth it to figure out the test framework and get this covered, or is this a one-and-done thing?

app/controllers/actors/ProvisionerProcessor.scala
nsauro: ec is implicit here.  If you want to specify implicit parameters in a scala function, the parameter list needs to begin with keyword 'implicit', and then every parameter in that list is implicit.

http://stackoverflow.com/questions/10635406/defining-a-function-with-multiple-implicit-arguments-in-scala

maddalab:
lgtm

maddalab:
I tested this today, I am going to go ahead and merge this, feel free to grab me a nick at work if required.

PR #265, opened 2015-02-15
removed finagle dep, and replaced SoftLayerPlugin's client with the Play...

nsauro:
... WS client


byxorna:
:runner: seems good to me, but I didnt test

maddalab:
:accept: +1

PR #267, opened 2015-02-16
Use configured threadpool

maddalab:
Remove the use of non configured thread pool which create threads
on the fly in favor on configured thread pool, which allows
for greater configuration flexibility

Builds on PR#266 and to be merged in only after that


app/collins/intake/IntakePlugin
```diff
@@ -1,37 +0,0 @@
-package collins.intake
```
Primer42: This seems important. Why are you deleting it?

conf/dev_base.conf
```diff
@@ -253,3 +253,57 @@ solr {
   externalUrl="http://localhost:8983/solr"
   embeddedSolrHome = "conf/solr/"
 }
+
+# Thread pool Configuration
+# ~~~~~
+
+# https://www.playframework.com/documentation/2.2.x/ThreadPools
+# these options would be the default for current collins nodes
```
Primer42: This info about our asset maybe doesn't belong here.

PR #268, opened 2015-02-17
Solr bug cleanup

Primer42:
There were a few issue that I found when I quickly looked at the solr code.
Firstly, it is not disabled by default as the docs imply.  In fact, it's always enabled.
Secondly, the default value for `connectionTimeout` was way too low, so I increased it to the default value recommended on the solr wiki.
Thirdly, there was a debug print that someone committed.
Once this is approved and merged, I'll update the default value in the docs accordingly.
@byxorna @maddalab


maddalab:
lgtm, requires a gh-pages update

app/collins/solr/SolrConfig.scala
```diff
@@ -16,17 +16,18 @@ object SolrConfig extends Configurable {
       messageWithDefault("invalidUrl", "solr.externalUrl %s is invalid".format(t), t)
   }
 
+  # defaults taken from http://wiki.apache.org/solr/Solrj#Changing_other_Connection_Settings
   def embeddedSolrHome = getString("embeddedSolrHome", "NONE")
-  def enabled = true
+  def enabled = getBoolean("enabled", false)
```
KirilAngov: Wouldn't we consider this a BC breaking change? Is the `true` defined somewhere else as the default?

byxorna: The solr plugin can not ever be disabled, so forcing this to true is ok (collins no worky without solr)

byxorna:
LGTM

app/collins/solr/SolrConfig.scala
KirilAngov: If @byxorna is right, we need to default it to true then: `def enabled = getBoolean("enabled", true)`

maddalab: I'm with @KirilAngov  on this, we should update the docs to say it is enabled by default.

app/collins/solr/SolrConfig.scala
```diff
@@ -16,17 +16,18 @@ object SolrConfig extends Configurable {
       messageWithDefault("invalidUrl", "solr.externalUrl %s is invalid".format(t), t)
   }
 
+  # defaults taken from http://wiki.apache.org/solr/Solrj#Changing_other_Connection_Settings
```
maddalab: # is not used for comments in scala

maddalab:
@Primer42 before we merge this we need updates to the reference config so that we do not forget

KirilAngov:
:+1:

PR #270, opened 2015-02-26
switched tests to use a console logger with a root of INFO.  This should...

nsauro:
... reduce logback spammage in tests


nsauro:
Just trying to fix logback in the test scope.  I switched to use a console logger, and this should reduce all of that awful log spam we get when running tests.

nsauro:
https://travis-ci.org/tumblr/collins/jobs/52287071

That is the output now

byxorna:
lgtm

maddalab:
lgtm

PR #272, opened 2015-03-02
When lshw output is missing capacity lets try size before resorting to the default

maddalab:


maddalab:
@nsauro @byxorna @Primer42

byxorna:
lgtm

defect:
lgtm too.

PR #275, opened 2015-03-13
Add id to log endpoints

Primer42:
Two changes
- When creating a log, it now returns the log entry, not just a boolean
- You can retrieve logs by id
I also added these abilities to `collins-client`

This will make the logging api much more useful, allowing programs to store and retrieve specific log messages, instead of asking an asset all of its logs, and hoping you can use regexes to find the one you need.

@nsauro @maddalab @byxorna @roymarantz @defect


maddalab:
no new tests?

Primer42:
@maddalab good call - tests added :smile:

defect:
Ruby changes LGTM. Don't forget to update the gh-pages docs. :)

Primer42:
Yeah, I'm going to do a big doc update once this is out.

byxorna:
LGTM bu i havent tested

PR #277, opened 2015-03-17
I've been using this to generate changelogs.  

Primer42:
Thought I might as well save it in the repo, so others can use it too.

@byxorna @maddalab @roymarantz @defect


maddalab:
lgtm

PR #279, opened 2015-03-20
[WIP] Remove the use of arbitrary dispatcher and use play built in actor system

maddalab:
Summary:
After our fumbles with ipmi errors, reviewed the implementation and use
of dispatcher. This refactoring makes proper use of plays actor system.
We can document the actors when they are used. The actors can be configured
using plays configuration (same as the previous attempt). Collins
now only requires the default-dispatcher on account of using
`Execution.Implicits` at a few places. The remainder of Collins relies on
actors

WIP - Need additional testing, basic smoke tests complete and succeed

@byxorna @defect @nsauro @DanSimon @Primer42

### DIFF
--- a/app/util/IpmiCommand.scala
+++ b/app/util/IpmiCommand.scala
@@ -72,7 +72,7 @@ abstract class IpmiCommand extends BackgroundProcess[Option[CommandResult]] {
     Some(cr)
   }
 
-  override protected def defaultTimeout: Duration = Duration(2, TimeUnit.MILLISECONDS)
+  override protected def defaultTimeout = Duration(2, TimeUnit.MILLISECONDS)
 
   protected def substitute(cmd: String): String = {
     cmd.replace("<host>", address)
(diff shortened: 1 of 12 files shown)


app/util/IpmiCommand.scala
```diff
@@ -72,7 +72,7 @@ abstract class IpmiCommand extends BackgroundProcess[Option[CommandResult]] {
     Some(cr)
   }
 
-  override protected def defaultTimeout: Duration = Duration(2, TimeUnit.MILLISECONDS)
```
byxorna: Really? This seems like a crazy small default timeout. Not related to this change, but...

byxorna:
LGTM! Out with the old.

:boom:

app/util/IpmiCommand.scala
maddalab: This should use the timeout in concurrency_reference.conf. It was overridden, we should remove it as part of some later PR. We may also want to break out the timeout for ipmi from the timeouts for other background operations at the time.

PR #266, opened 2015-02-16
Externalize akka configuration for background process and solr

maddalab:
Summary:
Externalizing akka configuration will enable updates to akka without
requiring code changes

Apache Bench - GET - All assets

➜  ~  ab -n 50000 -c 100 -A blake:admin:first http://localhost:9000/api/assets
This is ApacheBench, Version 2.3 <$Revision: 655654 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 5000 requests
Completed 10000 requests
Completed 15000 requests
Completed 20000 requests
Completed 25000 requests
Completed 30000 requests
Completed 35000 requests
Completed 40000 requests
Completed 45000 requests
Completed 50000 requests
Finished 50000 requests

Server Software:
Server Hostname:        localhost
Server Port:            9000

Document Path:          /api/assets
Document Length:        55627 bytes

Concurrency Level:      100
Time taken for tests:   222.486 seconds
Complete requests:      50000
Failed requests:        0
Write errors:           0
Total transferred:      2792700000 bytes
HTML transferred:       2781350000 bytes
Requests per second:    224.73 [#/sec] (mean)
Time per request:       444.972 [ms] (mean)
Time per request:       4.450 [ms] (mean, across all concurrent requests)
Transfer rate:          12258.05 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   3.2      0     208
Processing:    19  444 309.7    419    2945
Waiting:       18  444 309.7    418    2945
Total:         19  445 309.8    419    2945

Percentage of the requests served within a certain time (ms)
  50%    419
  66%    534
  75%    606
  80%    661
  90%    855
  95%   1022
  98%   1219
  99%   1395
 100%   2945 (longest request)

@nsauro @byxorna @DanSimon @Primer42

PR #281, opened 2015-03-20
Revert "Externalize akka configuration for background process and solr"

Primer42:
Reverts tumblr/collins#266


Primer42:
close in favor of #282

PR #283, opened 2015-03-22
Externalize and clean up usage of actor systems

maddalab:
Summary:
This attempt to do the same changes as the previous 3 commits
that we reverted in a single change set. The primary difference
between these changes and the ones reverted is the use of FromConfig
is the instantiation of router actors.

Why : Provides configurability of concurrency options by
removing these from code and moving them to a configuration file

Difference : Previously only the BackgroundProceesor actor
used a router actor. With this change even solr actors use a router
actor. In practice this means that we have had limited
concurrency for SOLR changes. A single actor was processing
the mail box for each SOLR message type (3 types). We can now 
have multiple actor processing the mailbox. See Consequences
below.

Non orthogonal changes:
Consolidation of adhoc use of threadpools to use the actor system.
Going forward we should not introduce any execution context,
implicit or otherwise and try to make use of the actors.

Breaking Changes:
The use of FromConfig necessitates updated configuration files,
users must specific at a minimum the router configuration. The
can further tune the additional parameters such as mailbox options,
timeouts, resizer settings etc [1]

Consequence:
Making sure SOLR options are appropriate. When testing this with
Solr configured for 128 instance, the default setting for solr
auto-warming were not adequate. This should be mentioned in the
release notes are requiring an update if using more than 1 routee.
Using 1 routee maintains existing behavior and does not need
and solr updates

[1] http://doc.akka.io/docs/akka/2.1.4/general/configuration.html


maddalab:
Now that this has been up and running for a few days without issues and since this is the consolidate PR of the 3 previously reviewed I am going ahead and merging it.

PR #284, opened 2015-03-26
Move all timeouts to external config

maddalab:
Summary:
Eliminating the use of hardcoded configs. The only timeout on in
config is TestProcessor which is a testing endpoint for api/asyncPing.

Please review the conf/* files including the reference files and
ActivationProcessor and AssetCancelProcessor which take config
values from SoftlayerPlugin. My reading of the code for Activation
and AssetCancel was that these were only used/enabled with
Softlayer.

@byxorna @Primer42 @defect @nsauro @DanSimon


app/controllers/actors/ActivationProcessor.scala
```diff
@@ -6,10 +6,11 @@ import play.api.mvc.{AnyContent, Request}
 import util.concurrent.BackgroundProcess
 import util.plugins.SoftLayer
 import scala.concurrent.{Await, Future}
+import java.util.concurrent.TimeUnit
+import collins.softlayer.SoftLayerConfig
 
 case class ActivationProcessor(slId: Long, userTimeout: Option[FiniteDuration] = None)(implicit req: Request[AnyContent]) extends BackgroundProcess[Boolean] {
-  override def defaultTimeout = 60 seconds
```
maddalab: I believe `ActivationProcessor` is only used with `Softlayer`

byxorna:
Seems reasonable to me. This needs lotsa docs to make it clear where all the timeouts are configured (for each plugin), and what their defaults are.

PR #289, opened 2015-04-07
Update permissions.yaml

funzoneq:
This status was missing, causing my curl for a user in the tools group to fail.

@Primer42 @jmackey @byxorna @defect @maddalab


Primer42:
LGTM

PR #291, opened 2015-04-26
Rewriting ldap authentication to be more flexible.

maddalab:
Summary:
More configuration parameters that allow for greater range of
ldap deployment variation for authentication.

Starting with [1] builds on it to complete the suggestion on that PR, specifically the second paragraph.

@asheepapart @byxorna @Primer42 @nsauro 

Tested with ldap authentication, these changes are backwards compatible and will not need any updates to config

[1] https://github.com/tumblr/collins/pull/100

### DIFF
--- a/app/util/security/LdapAuthenticationProvider.scala
+++ b/app/util/security/LdapAuthenticationProvider.scala
@@ -29,6 +29,8 @@ class LdapAuthenticationProvider extends AuthenticationProvider {
   def binddn = config.binddn
   def bindpwd = config.bindpwd
   def anonymous = config.anonymous
+  def groupnameattrib = config.groupNameAttribute
+  def usernumattrib = config.userNumberAttribute
 
   // setup for LDAP
   protected def env = Map(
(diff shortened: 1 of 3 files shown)


asheepapart:
@maddalab This is a great improvement. Thanks!

byxorna:
@roymarantz 
LGTM in a cursory pass

app/util/security/LdapAuthenticationProvider.scala
```diff
@@ -32,76 +38,116 @@ class LdapAuthenticationProvider() extends AuthenticationProvider {
...
-  protected def getSecurityPrincipal(username: String): String = {
-    "%s,%s".format(getPrincipal(username), searchbase)
-  }
-
-  // returns (&cn=*)(uniqueMember=uid=USERNAME,ou=people,dc=example,dc=org)
-  protected def groupQuery(username: String): String = {
+  protected def groupQuery(dn: String, username: String): String = {
     if (config.isRfc2307Bis)
-      "(&(cn=*)(%s=%s))".format(groupattrib, getSecurityPrincipal(username))
+      "(&(%s=*)(%s=%s))".format(groupnameattrib, groupattrib, dn)
```
roymarantz: dn seems like a funny attribute to use for a group lookup.  I would think basedn would be better, but I'm not sure what value it is likely to have.  Do you expect dn == 'ou=people,dc=example,dc=org' or 'ou=group,dc=example,dc=org' ?

roymarantz:
looks pretty good, but there is lots of magic that I didn't investigate closely.

PR #294, opened 2015-05-07
Get rid of Threadlocal user option

maddalab:
Summary:
Threadlocals are a poor mans globals and make just about as much sense
in a event based systems where thread can be put to use in different
context. The lack of calls to removeUser and the majority of uses
being `setUser(None)` is a good indication this isn't a good choice

@byxorna @nsauro @Primer42

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -23,7 +23,6 @@ object Application extends SecureWebController {
   )
 
   def login = Action { implicit req =>
-    setUser(None)
     req.queryString.get("location") match {
       case None =>
         Ok(html.login(loginForm))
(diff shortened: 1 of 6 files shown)


byxorna:
Not really sure what the purpose of `AppConfig.getUser` and friends were before. Seems like a vestigial bit of code?

From a layman's perspective, :cake:

maddalab:
SBT could not download a dependency, since there aren't any dependency changes, I',m closing a reopening this PR to create a new build

maddalab:
one of our repos used to provide the dependency and no longer does, I'm adding a new resolver to get it compiling, reproduced it locally by deleting restlet jar from play cache.

PR #296, opened 2015-05-17
Upgrading play to 226

maddalab:
This is good to review. I have described each commit below in addition to the commit messages which include `summary`.

### Reorganize structure of Collins

This commit isn't important from the standpoint of upgrading play 2.2.6 I did this to wrap my head around our inconsistent use of `package` and `import` directives.

```
package collins
package controllers
```
is the same as
```
package collins.controllers
import collins._
```
on account of scala's nested package scheme. This may very well be the standard across other projects, I have seen this used in [colossus] (https://github.com/tumblr/colossus/blob/master/colossus/src/main/scala/colossus/controller/Controller.scala) but it is very confusing, and I am resorting to using package and import directives as in Java. I looked up a internet documentation and could not find anything where there is consensus on how to do this right and what works best.

*Does not requires an exhaustive review and is also the largest changeset*

### Upgrade the build files to 2.2.6

As the commit states only deals with build files. Play cache module beginning with 2.2.6 is split into its own [module](https://www.playframework.com/documentation/2.2.x/Migration22) . Play organization has been updated to `com.typesafe`

### Handle the most basic of errors for 2.2.6

This big changes in play 2.2.6 that affect us are the "new result structure" and "actor composition". Alongside these, changes around `JsObject` hierarchy bring in smaller changes. This commit is the first of a series that brings about these changes

### Making the transition to 2.2.6

More of the same as above, this time a lot more exhaustively

### Address issues with Mixed Authentication provider

This isn't related to the upgrade. I noticed as part of working on the Upgrade issues with Mixed authentication provider. This addresses them

### Pushing the futures onto actions inheriting from `SecureAction`

`SecureAction` defined a method `execute` to be overridden by derived actions and return a `Result`. 

The type hierarchy of `Result` include `AsyncResult` which is effectively `Future[Result]`, since `Result` and `AsyncResult` are deprecated and scheduled for removal with `2.3.` releases of play, the earlier commits make the switch to `SimpleResult`. 

`AsyncResult` isn't a type of `SimpleResult` resulting in calls to `Await.result` in derived classes with fixed/hardcoded duration. 

Pushing down the the return type `Future[Result]` allows us to remove the hard coded duration and let the framework appropriately handle futures.

### Adding scala style plugin.

As the commit says, added the scala style plugin for later fixing of reported issues.


byxorna:
In some simple tests I got an issue with authentication (just adding assets with `scripts/populate.sh`): 
```
-> $ ./scripts/populate.sh 
....
ASSET
        ID      3
        CREATED 2015-05-17T19:14:28
        TAG     tumblrtag301
        STATUS  Incomplete
        DELETED null
        UPDATED null
        STATE   null
        TYPE    SERVER_NODE
IPMI
        IPMI_USERNAME   root
        ID      3
        IPMI_ADDRESS    172.16.32.21
        IPMI_NETMASK    255.255.240.0
        ASSET_ID        3
        ASSET_TAG       tumblrtag301
        IPMI_GATEWAY    172.16.32.1
        IPMI_PASSWORD   JeGe1ZmgTW48ect5
Invalid Username/Password specified^C
```

From the application logs:
```2015-05-17 19:14:34,212 - [WARN] - application - c.c.SecureApiController$class:authenticate:148 - Caught exception authenticating user: No authentication handler available```

maddalab:
Was the app reloading? I noticed in dev that the app reload frequently and I'm not sure why. @byxorna

byxorna:
@maddalab the incessant reloading happened on play 2.1.x as well (and 2.0.8 iirc) in dev mode as well. I think the play magic that detects changes to source code is seeing changes to files in the solr data directory and reloading all the time... Just a guess

maddalab:
You can run into a situation (timing based) when the app is reload that the authentication is not configured. I imported the assets using `populate.sh` ran thru intake and provisioning.

maddalab:
@nsauro @byxorna @Primer42

PR #297, opened 2015-05-19
Cleanup collins_state and make it ready for >1.9.2

defect:
This is my attempt at getting collins_state a bit more up to date. Currently, it will only work on 1.9.2 since the default value for the `:create_additions` parameter to `JSON.parse` changed from `true` to `false`.

The only actual code change is in 8d40f6c6ad44d19e8a741d21ec7a9f3888479deb. When `:create_additions` defaults to `false`, former line 130 (https://github.com/tumblr/collins/blob/master/support/ruby/collins-state/lib/collins/state/mixin.rb#L130) would return a `Hash`, and we would never convert it into a `Collins::State::Specification`. Instead, set `:create_additions` to `false` explicitly in all versions and let `json_create` take care of it.

I'm also cleaning up some tumblrisms by removing jeweler and references to internal systems,  as well as bumping a bunch of dependency versions and fixing various tests. I've intentionally kept the rspec version at < 3 since there's a lot of changes going on between 2.99 and 3.x.

I've run the test suite on 1.9.2, 1.9.3, 2.1 and 2.2 (with @byxorna's docker images) and they all pass.

@maddalab @Primer42 @byxorna @roymarantz


byxorna:
sweet! :cake: lgtm

maddalab:
lgtm

support/ruby/collins-state/lib/collins/state/mixin.rb
```diff
@@ -127,11 +127,11 @@ def state_specification asset
     updated = asset_from_cache asset
     result = updated.send(attribute_name.to_sym)
     if result then
-      res = JSON.parse(result) rescue nil
-      if res.nil? then # for backwards compatibility
-        res = JSON.parse(result, :create_additions => false)
+      res = JSON.parse(result, :create_additions => false) rescue nil
+      if not res.nil? then
         res = ::Collins::State::Specification.json_create(res) if (res.is_a?(Hash) and res.key?('data'))
```
roymarantz: any reason not to combine the previous line's conditional with this one since there is only this one line in the if/then?

roymarantz:
besides my comment :+1:

support/ruby/collins-state/lib/collins/state/mixin.rb
defect: Yeah, that should would work, and makes things a bit tidier.

support/ruby/collins-state/collins_state.gemspec
```diff
@@ -32,22 +27,17 @@ Gem::Specification.new do |s|
   s.homepage = "https://github.com/tumblr/collins/tree/master/support/ruby/collins-state"
   s.licenses = ["APL 2.0"]
   s.require_paths = ["lib"]
-  s.rubygems_version = "1.8.23"
   s.summary = "Collins based state management"
 
-  if s.respond_to? :specification_version then
-    s.specification_version = 3
+  s.add_runtime_dependency 'collins_client'
```
defect: Should we add some more pessimistic version dependencies maybe? At least have the runtime dependencies pinned to their current minor version?

roymarantz: IMO pin to what ever level indicates a the collins_client API has not changed.

support/ruby/collins-state/collins_state.gemspec
```diff
@@ -1,15 +1,11 @@
...
-# DO NOT EDIT THIS FILE DIRECTLY
-# Instead, edit Jeweler::Tasks in Rakefile, and run 'rake gemspec'
 # -*- encoding: utf-8 -*-
 
 Gem::Specification.new do |s|
   s.name = "collins_state"
   s.version = File.read 'VERSION'
 
-  s.required_rubygems_version = Gem::Requirement.new(">= 0") if s.respond_to? :required_rubygems_version=
   s.authors = ["Blake Matheny"]
```
Primer42: @defect you could add yourself to the author's list if you wanted :smile: :cat2: :dizzy:

support/ruby/collins-state/collins_state.gemspec
```diff
@@ -1,15 +1,11 @@
...
-  s.date = "2012-11-19"
+  s.date = "2015-05-20"
   s.description = "Provides basic framework for managing stateful processes with collins"
   s.email = "bmatheny@tumblr.com"
```
Primer42: We should probably make this a more useful email address, since I doubt anyone is checking Blake's email anymore.

Primer42:
LGTM :dizzy_face: :100: :black_large_square:

PR #298, opened 2015-05-21
This would always fail, because only the first command was sudo'd

Primer42:
@byxorna @maddalab @defect @roymarantz


byxorna:
Eh, LGTM.

PR #300, opened 2015-05-24
Update the metrics library

maddalab:
@nsauro @byxorna @Primer42 @DanSimon

### DIFF
--- a/app/Global.scala
+++ b/app/Global.scala
@@ -1,3 +1,5 @@
+import scala.concurrent.Future
+
 import play.api.Application
 import play.api.GlobalSettings
 import play.api.Logger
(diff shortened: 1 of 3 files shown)


app/Global.scala
```diff
@@ -18,12 +22,10 @@ import collins.util.config.CryptoConfig
 import collins.util.security.AuthenticationAccessor
 import collins.util.security.AuthenticationProvider
 import collins.util.security.AuthenticationProviderConfig
-import scala.concurrent.Future
-import play.api.mvc.SimpleResult
 
 object Global extends GlobalSettings with AuthenticationAccessor with CryptoAccessor {
   private[this] val logger = Logger.logger
-
+
```
Primer42: Can we clean up this whitespace? And add a newline back to the end of the file?

cburroughs:
metrics-reporter-config will also need to be updated when going from 2.x-->3.x

maddalab:
I discarded this commit, will redo it once I think of a better way, not convinced this was done right.

Issue #223, opened 2014-09-20
python client library?


ytjohn:
The document several times mentions a python client library available. I have been unable to locate this (only the ruby client). Is it publicly released?

cburroughs:
It looks like https://github.com/tumblr/collins/blob/master/support/python/collins_client/  has landed, but I don't see anything on pypi. Still a work in progress?

Primer42:
Sorry for the delays.  I've been working on this, and have been trying to track down the right creds/certs to host it on pypi.  I'll try to have it published this week.
That said, you can run from that library if you want.  The library isn't as fully functional as the ruby client, so if you find any bugs or make any improvements, we would love it if you could contribute them back!

Primer42:
Hi all.
Sorry for dragging my feet on this.
The python `collins-client` library should be available at `pypi` here: https://pypi.python.org/pypi/collins-client/0.1.0
Please let me know if you have any issues with it!

axlroden:
@Primer42 I just tried installing the pypi module.. 

Not sure what you are missing, but it's not actually installing the collins_client.py into site-packages on the OS ? all the dependencies etc are working ok if you copy the .py file in on your own..
The homepage/url seems wrong as well ..

defect:
I fixed the pypi package and pushed a version [0.1.1](https://pypi.python.org/pypi/collins-client/0.1.1). Please try it out and let us know if you have any issues.

I'm closing this issue, but feel free to reopen this or open a new ticket!

PR #301, opened 2015-05-25
Make python client an actual package

defect:
Just restructuring the python client more like a python package so that setup.py finds it. As noted in #223 installing the package from pypi doesn't work, and this should fix that.

@Primer42 @maddalab @byxorna


Primer42:
Thank you! LGTM

byxorna:
:cake:

PR #306, opened 2015-05-29
Update IpmiApi.scala

froxart:


maddalab:
@Primer42 @defect @byxorna

app/collins/controllers/IpmiApi.scala
```diff
@@ -41,7 +41,7 @@ trait IpmiApi {
   val IPMI_FORM = Form(
     mapping(
       "username" -> optional(text(1)),
-      "password" -> optional(text(8)),
+      "password" -> optional(text(minLength=5, maxLength=16)),
```
defect: My only comment is that in IPMI 2.0 the max password length is 20. Maybe bump this to that? I believe earlier IPMI versions (where the max was 16 chars) will truncate the password if its too long, but I'm not 100%.

Primer42: @defect yes on earlier versions of IPMI, it truncates to 16 chars if you pass it something too long.  I agree, let's increase this to 20 so it can support IPMI 2.0.

maddalab:
@froxart thanks for the pull request. i merged it.

froxart:
fantastic, thanks a lot for that fast response

PR #307, opened 2015-05-29
collins_client clean up

defect:
Basically doing the same thing i did for collins_state. Get rid of jeweler, cleanup the gemspec, bump a couple of gem versions. I've intentionally kept httparty at 1.11.0 since in later versions they drop support for ruby 1.9.2. This should work for the case in https://github.com/tumblr/collins/pull/204 since the gem conflicting there only requires httparty `~> 1.11.0`.

I'm also updating a bunch of tests to fix deprecations warnings. The tests run cleanly on 1.9.2, 1.9.3, and 2.2.

@byxorna @Primer42 @maddalab @roymarantz


support/ruby/collins-client/Gemfile.lock
```diff
@@ -1,50 +0,0 @@
-GEM
```
byxorna: We should keep the gemfile.lock in versioncontrol, and just run a `bundle update`. That way we know the exact versions of things that work when we are devving.

defect: If we want exact versions of gems we depend on I think those should go in the gemspec. I can add some pessimistic version requirements to the gemspec based on what I've been using when testing it on the various versions of ruby.

roymarantz:
:+1:  if this really works :-)

support/ruby/collins-client/Gemfile.lock
Primer42: I agree with @defect - for gems you should not have Gemfile.lock checked in.

support/ruby/collins-client/collins_client.gemspec
```diff
@@ -1,8 +1,3 @@
-# Generated by jeweler
-# DO NOT EDIT THIS FILE DIRECTLY
-# Instead, edit Jeweler::Tasks in Rakefile, and run 'rake gemspec'
-# -*- encoding: utf-8 -*-
-
 Gem::Specification.new do |s|
   s.name = "collins_client"
   s.version = File.read 'VERSION'
```
Primer42: I think you need to do a version bump still

Primer42:
LGTM :ship: :it:



(the ship emojii, and then the it emoji...... get it?)

PR #310, opened 2015-06-02
Fix collins-notify for ruby 2.2.0

byxorna:
@maddalab @Primer42 @roymarantz @defect This fixes the `#respond_to?` invocations to actually find protected methods, which is preventing collins-notify from working on newer rubies. We ask respond_to to search all methods, not just public ones.

[0] http://ruby-doc.org/core-1.9.3/Object.html#method-i-respond_to-3F
[1] http://ruby-doc.org/core-2.2.2/Object.html#method-i-respond_to-3F

I have tested this and confirmed functionality in both 2.2.0 and 1.9.2.


support/ruby/collins-notify/collins_notify.gemspec
```diff
@@ -44,28 +44,10 @@ Gem::Specification.new do |s|
...
-    s.add_dependency(%q<collins_client>, ["~> 0.2.10"])
-    s.add_dependency(%q<hipchat>, ["~> 0.7.0"])
-    s.add_dependency(%q<mail>, ["~> 2.4.4"])
-    s.add_dependency(%q<nokogiri>, ["~> 1.5.2"])
-  end
+  s.add_runtime_dependency('addressable', '~> 2.3.2')
+  s.add_runtime_dependency('collins_client', '~> 0.2.10')
+  s.add_runtime_dependency('hipchat', '~> 0.7.0')
+  s.add_runtime_dependency('mail', '~> 2.4.4')
+  s.add_runtime_dependency('nokogiri', '~> 1.5.2')
```
defect: Maybe move the dev dependencies in here too and just point the Gemfile to the gemspec?

byxorna: fixed.

defect:
lgtm

roymarantz:
:+1:

PR #314, opened 2015-06-08
Fix the error message displayed when login fails

maddalab:
Summary:
Exception handling to better report a login failure and not display
an exception code.

@Primer42 @byxorna @defect

### DIFF
--- a/app/collins/util/security/LdapAuthenticationProvider.scala
+++ b/app/collins/util/security/LdapAuthenticationProvider.scala
@@ -96,38 +96,50 @@ class LdapAuthenticationProvider extends AuthenticationProvider {
   }
   
   // creating a initial dir context will fail to indicate an authentication error
-  private def getUserContext(dn: String,  password: String) = {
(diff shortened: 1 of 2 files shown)


app/collins/util/security/LdapAuthenticationProvider.scala
```diff
@@ -96,38 +96,50 @@ class LdapAuthenticationProvider extends AuthenticationProvider {
...
-      } catch {
-        case e: AuthenticationException =>
-          logger.info("Failed authentication for user %s", e)
-          None        
-      } finally {
-    	ctx.close
-      }
+      })
+    } { t =>
+      logger.error("Failed to bind to ldap server, please verify ldap configuration")
```
byxorna: Worthwhile spitting out the error here?

byxorna:
LGTM, not tested though. Have you verified for both local and LDAP accounts?

maddalab:
verified with file and ldap auth

PR #318, opened 2015-06-11
ghe pages formatting overhaul

Primer42:
This redoes how the documentation is organized.  It:
- Make the docs much more DRY, automatically generated menus from YML files and keeping as much formatting in templates as possible
- Upgrades to bootstrap 3

After this PR goes through, I'll be going through the docs and updating their content, rather than their structure.
@byxorna @maddalab @defect @roymarantz


byxorna:
@Primer42 fuck yeah! 
screen shot 2015-06-11 at 4 28 31 pm

No way in hell this reviewable via github. Ill check it out and play with it.

Primer42:
Yeah I was hoping it would be a net negative.  But I think going to bootstrap 3 messed up my stats.

cburroughs:
I'm not how feasible this is with GH's render magic but it would be nice if the docs were in `master` so that PRs could change the docs at the same time.

Primer42:
@cburroughs that would be nice, but it would mean we would have to depart from GH's magic. I've thought about it, and I can't think of a nice way to do it.  But I'm open to suggestions :smile:

roymarantz:
Can't some of the yaml files be generated from the sources?  At least the API calls seem like that should be possible.  Or is that too much like literate programming :-)

[I actually looked (briefly) at all 164 files hoping to see something like that]

Primer42:
@roymarantz totally - I think that's the next step, especially for the API.  Or we just use a tool that is built to generate api docs like this.

maddalab:
make sure the help link from the application link to the correct locations on the new pages. [example] (http://localhost:9000/help?t=default#assetState)

Primer42:
@maddalab I'll open a new PR that fixes things on the application side

maddalab:
ok, lgtm

PR #319, opened 2015-06-13
Remove use of async result during asset cancel request

maddalab:
Summary:
Remove deprecated API AsyncResult and replace with
Action. async when canceling an asset.

I tested intake and provisioning, however cancel itself requires softlayer and the particular change has not been tested. Given the change, I'm confident it is ok.


byxorna:
@maddalab can you write some tests for this, maybe stub out the softlayer stuff? Looks simple enough, but I'd feel better if we made a habit of adding tests, especially where we have no good way to verify functionality

maddalab:
I added a note in the issue around replacing softlayer api with jclouds to add tests. Since we plan to do that for the next release and I need this change in a sequence of steps to implement ipv6 ipam, going to go ahead and merge.

PR #320, opened 2015-06-14
Minor: A Two-Tuple of Options where only one of the tuple elements is Some at a time is an either

maddalab:
Summary:
Updating SendType from (Option[Throwable], Option[T]) to
Either[Throwable, T]. Only of the options in the tuple
is Some. This is better represented as an Either

@nsauro @DanSimon @byxorna @Primer42


byxorna:
LGTM on the surface. Much more understandable!

PR #321, opened 2015-06-14
Minor: Upgrade solr and httpcomponents version

maddalab:
@byxorna @Primer42


byxorna:
What do we get out of this upgrade? Any specific features we want?

maddalab:
no idea what the new features are. Nothing more than upgrading the deps and ensure we can continue building.

PR #322, opened 2015-06-16
Update play to 2.3.9

maddalab:
The [migration guide] (https://www.playframework.com/documentation/2.3.x/Migration23) lists the changes. This isn't as big a change set as some of the other upgrades.

I switched to using [webjars] (http://webjars.org) and committed files for jquery / bootstrap etc are now removed. For funsies I switched the these to cerulean, which we can remove if we don't like it.

For reverse routing, I included the `webjars-play` package. Good introduction on [webjars] (https://www.youtube.com/watch?v=C4Lyx_d9JKI)

On `nl.grons.metrics-scala` we went back a version due to a bug in `2.2.0`

Tested: standalone, docker and packing scripts. API invocation was spot checked using the gem.

@byxorna @defect @Primer42 @roymarantz


maddalab:
Going to go ahead and merge this since it has been tested on the standby node and running without issues. The standby node on which I tried a few themes, is running `yeti` will open a follow up PR to switch that.

Issue #323, opened 2015-06-17
LSHW on KVM doesn't match 'bank'


yafsn:
I found that lshw within a KVM vm is reporting the node id for memory as 'bank' instead of 'bank:X'.   As a test I was able to update line 86 of [LshwParser.scala](https://github.com/tumblr/collins/blob/master/app/collins/util/parsers/LshwParser.scala) to match on simply 'bank', but I'm not sure if there is a better way to resolve this.  The end result without it is simply no memory listed for the asset.

```
# lshw -version
B.02.16
# lshw -c memory -xml
<?xml version="1.0" standalone="yes" ?>
<!-- generated by lshw-B.02.16 -->
<!-- GCC 4.8.2 -->
<!-- Linux 3.13.0-45-generic #74-Ubuntu SMP Tue Jan 13 19:36:28 UTC 2015 x86_64 -->
<!-- GNU libc 2 (glibc 2.19) -->
<list>
  <node id="firmware" claimed="true" class="memory" handle="">
   <description>BIOS</description>
   <vendor>Bochs</vendor>
   <physid>0</physid>
   <version>Bochs</version>
   <date>01/01/2007</date>
   <size units="bytes">98304</size>
  </node>
  <node id="memory" claimed="true" class="memory" handle="DMI:1000">
   <description>System Memory</description>
   <physid>1000</physid>
   <size units="bytes">2146435072</size>
   <capacity units="bytes">2146435072</capacity>
    <node id="bank" claimed="true" class="memory" handle="DMI:1100">
     <description>DIMM RAM</description>
     <physid>0</physid>
     <slot>DIMM 0</slot>
     <size units="bytes">2146435072</size>
     <width units="bits">64</width>
    </node>
  </node>
</list>
```

maddalab:
That fix works. Do you want to open a PR?

yafsn:
No problem.

PR #324, opened 2015-06-17
LSHW on KVM doesn't match 'bank'

yafsn:
Lshw within a KVM vm is reporting the node id for memory as 'bank' instead of 'bank:X'.  This change simply matches on 'bank' instead of 'bank:'.

ref: [bug-323](https://github.com/tumblr/collins/issues/323)

### DIFF
--- a/app/collins/util/parsers/LshwParser.scala
+++ b/app/collins/util/parsers/LshwParser.scala
@@ -83,7 +83,7 @@ class LshwParser(txt: String) extends CommonParser[LshwRepresentation](txt) {
   }
 
   val memMatcher: PartialFunction[NodeSeq,Memory] = {
-    case n if (n \ "@class" text) == "memory" && (n \ "@id" text).contains("bank:") =>
+    case n if (n \ "@class" text) == "memory" && (n \ "@id" text).contains("bank") =>
       val asset = getAsset(n)
       val size = (n \ "size" text) match {
         case n if n.isEmpty => ByteStorageUnit(0)


yafsn:
Its interesting where that failed.  It wasn't even referencing anything with 'bank' in lshw.

defect:
I think it might be picking up the BIOS flash memory? From the build output: `Memory(4.00 MB,12,FLASH Non-volatile 33 MHz (30.3 ns),26DF321,ATMEL`

https://github.com/tumblr/collins/blob/master/test/resources/lshw-old-web.xml#L411

yafsn:
That makes more sense @defect.  Before this memory wasn't being reported, so now it is and the higher count is throwing off the test.  It looks like the ':' was unintentionally avoiding that issue.  Adding an exception for 'slot == bios' seems hokey to me since this is the only place that's seen (so far).

defect:
Yeah, i agree. Not quite sure how to fix it in a better way. Maybe check if the node has `claimed = true`? Reading [0] it looks like anything that the OS might actually be able to use would have that flag.

@maddalab @Primer42 @byxorna @roymarantz Thoughts?

[0] http://www.ezix.org/project/wiki/HardwareLiSter#Howtointerpretlshwsoutput

maddalab:
@defect felix, what version of lshw do we use with genesis at our dc?

defect:
I believe we have B.02.17 (latest stable) in our repo.

maddalab:
@yafsn Want to see if that version helps out. I am not sure we have a good fix here, at Tumblr we have spoken about writing a new end point with defined schema requirements to do a system report. We have also been bitten by lshw breaking backwards compatibility between versions. Also, if you can provide us with the full report, I'll take a look to see if there is something on there that we can use to distinguish report for vms.

maddalab:
@defect Can we use [slot] (https://github.com/tumblr/collins/blob/master/test/resources/lshw-old-web.xml#L416) to skip it?

Scheme for lshw is also available [here] (http://www.ezix.org/project/attachment/wiki/HardwareLiSter/lshw.xsd)

yafsn:
Wow.  With the latest lshw it doesn't report *any* hardware details in collins.  I've put the full `lshw -xml` in gists: [b-02-16](https://gist.github.com/yafsn/009db5915f6dc1446374#file-lshw-b-02-16-xml), [b-02-17](https://gist.github.com/yafsn/009db5915f6dc1446374#file-lshw-b-02-17-xml).

I can really see why you'd want an end point you can control.

maddalab:
Yikes memory is reported entirely differently

```xml
    <node id="memory" claimed="true" class="memory" handle="DMI:1000">
     <description>System Memory</description>
     <physid>1000</physid>
     <size units="bytes">2146435072</size>
     <capacity units="bytes">2146435072</capacity>
      <node id="bank" claimed="true" class="memory" handle="DMI:1100">
       <description>DIMM RAM</description>
       <physid>0</physid>
       <slot>DIMM 0</slot>
       <size units="bytes">2146435072</size>
       <width units="bits">64</width>
      </node>
    </node>
```

vs

```xml
    <node id="memory" claimed="true" class="memory" handle="">
     <description>System memory</description>
     <physid>0</physid>
     <size units="bytes">2098221056</size>
    </node>
```

Based on the size, I'm assuming this is a different machine, right? 

However I do not have a good answer for you regardless, I leaning toward defining a Collins hardware report schema and requiring the report posted meets the schema requirement. We will define a more strongly types schema than the one I linked above from lshw but the schema will support the output from version of lshw we use in house at Tumblr today.

This will move additional burden of generating conforming output from hardware report onto users.

Keith, I will look into the possibility of using `slot` to distinguish, and if that does not work, I'd recommend you stay with `02-16` and implement a `xslt` transformation for `bank` to `bank:X`

yafsn:
Sounds good @maddalab.  I'll revert lshw so I can get some data in there for now and work on the xslt.  I'll close out this PR since it would be creating more issues than it fixes.

maddalab:
Thank Keith. Really appreciate your time and effort on this.

PR #325, opened 2015-06-18
Minor: Styling using bootswatch themes

maddalab:
Clean up css removing unused style specifications, unnecessary overrides etc. For live demo use the internal standby node.

@defect @byxorna @Primer42 @roymarantz


maddalab:
I'll go ahead and merge this because the primary changes are unused files, styles etc which mess with the base css styles. The selection of theme (sandstone - now) we can switch at any time to something else (including one we build)

PR #331, opened 2015-06-23
Make authentication logging less verbose

Primer42:
Once collins is used as a certain rate, this is all that appears
in the logs, and it makes them essentially useless.

@maddalab @byxorna @defect @roymarantz


roymarantz:
How about putting the authentication stuff in a separate log?  I suspect we don't care about it that often and it will be available, but out of the way then.

maddalab:
So, this is not right. Why would this logging get more verbose, when it has always existed.

PR #332, opened 2015-06-23
Use the tryAuthCache method to avoid making auth queries continously.

maddalab:
See #331 

Why would logging of backend auth become verbose for no apparent reason? Probably because we are instantiating a [new] (https://github.com/tumblr/collins/blob/master/app/collins/util/security/MixedAuthenticationProvider.scala#L44) provider for ever login request which is effectively busting the cache.

This attempts to address that along with addressing some software design issues. 

@Primer42 @defect @byxorna @roymarantz

### DIFF
--- a/app/collins/util/security/MixedAuthenticationProvider.scala
+++ b/app/collins/util/security/MixedAuthenticationProvider.scala
@@ -1,6 +1,12 @@
 package collins.util.security
 
+import java.util.concurrent.TimeUnit
+
 import collins.models.User
+
+import com.google.common.cache.CacheBuilder
+import com.google.common.cache.CacheLoader
+import com.google.common.cache.LoadingCache
 
 /**
  * Exception encountered during authentication phase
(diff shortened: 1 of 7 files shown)


roymarantz:
seems safe enough.  unit test?
:+1:

maddalab:
existing unit tests cover the area. However I am making an additional change.

app/collins/util/security/MixedAuthenticationProvider.scala
```diff
@@ -1,21 +1,70 @@
...
- * Provides authentication by a variety of methods
+ * Provides authentication by a variety of methods and caching logic (implements decorator pattern)
  */
 class MixedAuthenticationProvider(types: Array[String]) extends AuthenticationProvider {
-
-  /**
-   * @return The authorization type provided by this class
-   */
+  
+  private val providers = types.map({
```
maddalab: Instantiating the providers once will avoid the cache busting.

app/collins/util/security/MixedAuthenticationProvider.scala
```diff
@@ -1,21 +1,70 @@
...
+import com.google.common.cache.LoadingCache
+
 /**
  * Exception encountered during authentication phase
  */
 class AuthenticationException(msg: String) extends Exception(msg)
 
 /**
- * Provides authentication by a variety of methods
+ * Provides authentication by a variety of methods and caching logic (implements decorator pattern)
```
maddalab: Auth caching is an orthogonal concern and should not seep out to the rest of the code (i.e there should be no calls to `tryAuthCache` outside of this class)

app/collins/util/security/MixedAuthenticationProvider.scala
```diff
@@ -24,31 +73,6 @@ class MixedAuthenticationProvider(types: Array[String]) extends AuthenticationPr
...
-        val ldapProvider = new LdapAuthenticationProvider()
-        val user = ldapProvider.authenticate(username, password)
-        logger.debug("Tried ldap authentication for %s, got back %s".format(username, user))
-        user
-      }
-      case t => {
-        throw new AuthenticationException("Invalid authentication type provided: " + t)
-      }
-    }).headOption
+    providers.flatMap { p => tryAuthCache(p, username, password) }.headOption
```
maddalab: Needs update: Make lazy before comitting

roymarantz:
I don't know if you want to modify the old code, but you could probably reuse the connection to the ldap server instead of getting a new one each time.  Yes it will be messier code, dealing with retrying the connection on failure, but it would save the overhead of a new connection each time.

maddalab:
Which connection can be reused? The only instance we can reuse the connection is the initial context when ldap server allows for anonymous binds.

roymarantz:
Yes.  AFAIK, that is the only ldap connection we do. i.e. just connected
and then authenticate.

PR #334, opened 2015-06-23
cleanup the dockerfile some

byxorna:
Running `CMD` in the shell format (bare words) makes docker launch the thing in `sh -c "$CMD"`, which wont pass signals. Ive cleaned it up so we dont need env vars expanded when starting up, which means we can use the cleaner CMD format (sorry, Dockerfiles dont support multiline json arrays iirc)... Play is pid 1 inside the container, instead of pid 5 or so.

@Primer42 @maddalab @defect

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,4 @@
 _site
 target
 support
+vendor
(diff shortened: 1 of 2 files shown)


maddalab:
lgtm

Primer42:
LGTM

PR #336, opened 2015-06-27
Update README.md to use correct GitHub Pages URL

andrewjkerr:
Super small README change to update links from http://tumblr.github.com to the new GitHub Pages URL of https://tumblr.github.io. http://tumblr.github.com already redirects to https://tumblr.github.io so the links might as well be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,10 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


byxorna:
Thanks! LGTM

PR #337, opened 2015-07-03
Rethink use of Guava cache using play's plugin architecture

maddalab:
Summary:
What started off as a simple rewrite of tattler (logging user id) ran into issues
when I found ~20 unit tests that were not running.

Getting the unit tests to run resulted in failure and
so began a journey into the guts of the cache plugin architecture

Previously we use squeryl callback mechanism to implement
domain model caching. Why? I'm no so sure, but this did lead
to some fairly tight coupling of domain objects and caching logic.

We also use Guava cache at a few different places

(a) Cache file (user spec) for FileAuthentication
(b) Cachining permissions file
(c) Authenticated ldap users
(d) Metrics used with graphing
(e) Provisioning profiles

This pulls out the use of Guava cache from play's plugin arch
keeps the use cases other than domain objects in place, and
no longer caches domain objects.

Since we no longer use caching for domain objects, I updated 
the admin page to not display cache stats. We can always
add those back in.


maddalab:
And the entire reason to do this, 15 more unit tests
```
[info] Passed: Total 315, Failed 0, Errors 0, Passed 315
[success] Total time: 280 s, completed Jul 3, 2015 2:24:07 AM
```

maddalab:
Also as part of this PR, I made a change to allow for complete specification of Guava cache settings as provided for by [Guava CacheBuilderSpec](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/cache/CacheBuilderSpec.html).

maddalab:

PR #338, opened 2015-07-03
Fix an issue where tattler was not reporting the correct user for changes

maddalab:
Previously tattler log source selection was arbitrary, i.e. depends
    on the type of tattler in use. An api call can use user source or
    api source for tattling. Provide a standard means that ensures
    correctness of tattler source (By use SecureController).

    Secure Controller is aware of whether it is operating as Web or
    API. In addition, capture the user causing the tattle in a dedicated
    field on asset log instead of as part of the message. (Schema change)

    Also updated the log views to include the user field and
    formatted the date a little better.

This PR is into `bhaskar-cache-ectomy` as it builds off of changes on that.


maddalab:
image

maddalab:

maddalab:
merged into master, closing pr

PR #339, opened 2015-07-06
Do not use plugins when not extending the framework

maddalab:
Summary:
Previously we use play plugins to manage lifecycle of
components. We relied on the framework to return the
plugin via a reflective invocation to invoke methods on
these.

This change makes preparation for a play 2.4 upgrade.
Lifecycle is 2.4 is management by the the DI container.

Components of this change include

(a) Extract interfaces that can be mocked for testing and injected
(b) Trait companion objects for defaults implementations
(c) Remove the use of play plugin retrieval code

==================================================
Some scope creep that occurred as part of this change
is a rewrite of the generic frame implementation, which
was used by the Monitoring plugin.

The change implements the use of javascript call back
to determine availability of the frame and supports
arbitrary frame, which can be used for any asset type.

Example:

In dev base config to view Giphy frames with Tumblr
featured content.
```hocon
views {

  enabled = true

  frames = {
    giphy {
      enabled = true
      title = "Giphy"
      style = "width: 100%;height: 1200px;"
      script = """
        function isEnabled(asset) {
          return true;
        }

        function getUrl(asset) {
          return "http://giphy.com/search/tumblr-feat";
        }
      """
    }
  }
}
```
Implementing the monitoring frame
```hocon
views {
  enabled = true
  frames = {
    monitoring {
      enabled = true
      title = "Monitoring"
      style = "width: 100%;height: 1200px;"
      script = """
        function isEnabled(inp) {
          var obj = JSON.parse(inp);
          return ((obj.ASSET.TYPE === "SERVER_NODE") && ("HOSTNAME" in obj.ATTRIBS[0]));
        }
        function getUrl(inp) {
          var obj = JSON.parse(inp);
          return "https://icinga2.ewr01.tumblr.net/thruk/cgi-bin/status.cgi?host=" + obj.ATTRIBS[0]["HOSTNAME"];
        }
      """
    }
  }
}```


maddalab:

maddalab:
merged into master, closing pr

PR #343, opened 2015-07-09
fixed input search, was returning all results

MaximeDevalland:
I forgot the name tag. It was returning all results every time.
Now working. Sorry about that.


maddalab:
ty

PR #344, opened 2015-07-10
Changes to search bar

maddalab:
Some improvements to the search bar.

(a) Search can be done either using a CQL or tag
(b) Some minor changes to the header (search bar is bigger). Feature header items are to the left, only help and logout are on the right

Forms formatter (the code that converts) a string to a CQLQuery or SolrExpression (in the controller layer, not the CQL parser), now attempt to create a valid query string if the user provided string does not parse. This is done in exactly 1 case, if the provided string passes the asset tag constrains.

i.e. searching for `A1000` in the search bar or the home landing page text area will return the asset instead of returning an error.

image

@MaximeDevalland @Primer42 @defect @byxorna 

Maxime, while the asset tag was good, I felt it did not provide as much utility as it could with CQL. This will support the tag search in additional to CQL, thoughts?

### DIFF
--- a/app/collins/controllers/forms.scala
+++ b/app/collins/controllers/forms.scala
@@ -6,6 +6,7 @@ import play.api.data.FormError
 import play.api.data.format.Formats
 import play.api.data.format.Formatter
 
+import collins.models.Asset
 import collins.models.AssetSort
 import collins.models.AssetType
 import collins.models.State
@@ -15,6 +16,7 @@ import collins.power.PowerAction
 import collins.solr.AssetDocType
 import collins.solr.CollinsQueryParser
 import collins.solr.SolrExpression
+import collins.solr.CQLQuery
 
 package object forms {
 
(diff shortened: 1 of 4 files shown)


byxorna:
Nice! I would suggest adding a `?` icon next to the search box to punt you over to some CQL docs. Also, can the search bar be smaller, right justified?

I think a better ordering is 
```
Collins    Searchbar (?cqlhelp)                     Create Logging Logout (?)  
```

Or maybe even better, move the create/logging/logout into a hamburger menu? How does the searchbar fare when the display width is phone-sized? (when the buttons on the right fall into the hamburger menu)

maddalab:
A smaller search bar (width) does not work for CQL queries which can get long, similar to the browser search bar. Since none of us are design experts, I'll reach out internally to get a review later. I'll add a link to the help

maddalab:
image

Tried a few different options, nothing that I like and worked for different window sizes.. I'll be glad for someone else to try their hand at it. Not my forte, I', not updating the styles atm.

byxorna:
Seems ok to me. We will need to tweak it some more for sure

PR #348, opened 2015-07-21
Ensure only 1 instance of auth provider is ever created

maddalab:
check on provider types has been failing causing a new instance
to be created every time. Switching to List (immutable) addresses
the issue.

Summary:
scala> List("ldap") == List("ldap")
res0: Boolean = true

scala> List("ldap") != List("ldap")
res1: Boolean = false

scala> Array("ldap") != Array("ldap")
res2: Boolean = true

scala> Array("ldap") == Array("ldap")
res3: Boolean = false

### DIFF
--- a/app/collins/util/security/AuthenticationProvider.scala
+++ b/app/collins/util/security/AuthenticationProvider.scala
@@ -8,7 +8,7 @@ import collins.cache.GuavaCacheFactory
 
 trait AuthenticationProvider {
   protected val logger = Logger.logger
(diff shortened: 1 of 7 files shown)


roymarantz:
on the surface this looks good to me.
I would suggest you write a unit test to to test if == is working as you want since that is the motivator for this PR.

maddalab:
@roymarantz I don't want to write an eq test for a library class. I added a test for Global Authentication accessor

PR #349, opened 2015-07-21
api delete endpoint

gtorre:
@maddalab @Primer42


Primer42:
LGTM
To get a faster approval, I'm adding @roymarantz and @defect

support/ruby/collins-shell/lib/collins_shell/asset.rb
```diff
@@ -27,13 +27,15 @@ def create
       end
     end
 
-    desc 'delete', 'delete an asset in collins (must be cancelled)'
+    desc 'delete [NUKE=BOOLEAN]', 'delete an asset in collins (must be cancelled)'
```
roymarantz: What does "(must be cancelled)" mean?  I don't see any code which checks if the asset is in cancelled state or not.  Shouldn't this at least make sure the node isn't Allocated?

roymarantz:
:+1: 
this code is painful for me to read, but that is scale's fault not yours :-(

PR #351, opened 2015-07-27
Fixing relative paths and missing bracket

funzoneq:
@maddalab fixes. Also we might want to consider adding a "rubocop ." run to the automated testing. As a style guide and such.

### DIFF
--- a/support/ruby/consolr/lib/consolr.rb
+++ b/support/ruby/consolr/lib/consolr.rb
@@ -1,6 +1,9 @@
+$LOAD_PATH.unshift('./lib/')
+
 require 'collins_auth'
 require 'optparse'
 require 'yaml'
+require 'consolr/version'
 
 module Consolr
   class Console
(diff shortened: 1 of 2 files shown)


support/ruby/consolr/lib/consolr.rb
```diff
@@ -142,36 +145,27 @@ def start
...
       end
-      
+
       case
       when options[:console]
-        puts "--> Opening SOL session (type ~~. to quit)"
-        puts self.ipmitool_cmd("sol activate")
-      
+        puts '--> Opening SOL session (type ~~. to quit)'
+        puts ipmitool_cmd('sol activate')
```
roymarantz: You might like:
puts '--> Opening SOL session (type ~~. to quit)', ipmitool_cmd('sol activate')
or even:
puts '--> Opening SOL session (type ~~. to quit)',
   ipmitool_cmd('sol activate')

roymarantz:
:+1:  even with my comment

PR #352, opened 2015-07-29
Handle leading/trailing white space in top search bar field.

maddalab:
Summary:
Leading/trailing white space in top search bar field resulted in
"unexpected error occurred", trim the white space. Additionally
display a correct error message when the query in invalid.


roymarantz:
:+1:  but having a unit test would be better

Primer42:
LGTM I agree with @roymarantz

maddalab:
i added a test.

PR #353, opened 2015-08-04
Only fetch the required fields from solr.

maddalab:
Summary:
See AssetSearchQuery and AssetLogSearchQuery, we only rely on specific
fields. Fetch only these fields and not the entire document.

@byxorna @Primer42 @roymarantz 

TODO: Needs some real world testing.


byxorna:
Beautiful! Ill test this out later. :eyes: good

PR #354, opened 2015-08-05
Adding additional stats.

maddalab:
Summary:
Adding additional stats to help diagnose issues (if any). Grabbing solr fetch stats by page size (using arbitrary pages will increase the number of stats that would be collected).

@byxorna @defect @roymarantz @Primer42


Primer42:
LGTM

PR #356, opened 2015-08-07
Instrumenting with stats around the serialization of json.

maddalab:
Summary:
The previous change instrumented stats around the conversion
of data type to json objects. This reverts that change and
instrument the "stringification" of json.


roymarantz:
what are you doing this?

PR #358, opened 2015-08-11
[Experimental] Adds firehose functionality using websockets

maddalab:


roymarantz:
Please add a description including what this is supposed to do and why it is being done, FTR.
Also, shouldn't this get some tests added for this new functionality?

PR #359, opened 2015-08-14
Consolr dangerous asset behavior and unit tests

Primer42:
Fixed some bugs around how dangerous assets and statuses behave, and added unit tests.

@maddalab @byxorna @roymarantz @defect @sushruta @funzoneq

### DIFF
--- a/support/ruby/consolr/.gitignore
+++ b/support/ruby/consolr/.gitignore
@@ -1,0 +1,1 @@
+coverage/
--- a/support/ruby/consolr/spec/configs/collins_rspec.yml
+++ b/support/ruby/consolr/spec/configs/collins_rspec.yml
@@ -1,0 +1,4 @@
+host: https://collins.example.net
+username: "collins_user"
+password: "collins_pass"
+timeout: 30
--- a/support/ruby/consolr/spec/consolr_spec.rb
+++ b/support/ruby/consolr/spec/consolr_spec.rb
@@ -1,0 +1,133 @@
+require 'spec_helper'
+
+describe Consolr::Console do
+
+  before (:each) do
+    @console = Consolr::Console.new
+  end
+
+  it 'fails if no hostname is provided' do
+    expect { @console.start( {:hostname => nil}) }.to raise_error(SystemExit, /asset tag or hostname/)
+  end
+
+  it 'fails if both tag and hostname are passed' do
+    expect { @console.start({:hostname => 'host.dc.net', :tag => '001234' }) }.to raise_error(SystemExit, /not both/)
+  end
+
+  it 'fails if tag cannot be found' do
+    expect { @console.start({:tag => 'bogus_tag'}) }.to raise_error(SystemExit, /Found 0 assets/)
+  end
+
+  it 'fails if hostname cannot be found' do
+    expect { @console.start({:hostname => 'bogus_hostname'}) }.to raise_error(SystemExit, /Found 0 assets/)
+  end
+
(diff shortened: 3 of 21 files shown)


support/ruby/consolr/.gitignore
```diff
@@ -0,0 +1 @@
+coverage/
```
roymarantz: add a new line please

Primer42: Yep, was in the  middle of fixing it when you commented.

support/ruby/consolr/spec/consolr_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+
+    end
+
+    it 'does log list' do
+      expect { @console.start({:tag => dangerous_maintenance_tag, :log => 'list'}) }.to output("sel list\nSUCCESS\n").to_stdout_from_any_process
+    end
+
+    it 'does log clear' do
+      expect { @console.start({:tag => dangerous_maintenance_tag, :log => 'clear'}) }.to output("sel clear\nSUCCESS\n").to_stdout_from_any_process
+    end
```
roymarantz: how come these log commands are considered safe and tested via that mechanism?

support/ruby/consolr/spec/consolr_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+    expect { @console.start({:hostname => 'hostname-with-multiple-assets.dc.net'}) }.to raise_error(SystemExit, /Found \d+ assets/)
+  end
+
+  safe_boolean_actions = {:console => "--> Opening SOL session (type ~~. to quit)\nsol activate", :kick => 'sol deactivate', :identify => 'chassis identify', :sdr => 'sdr elist all', :on => 'power on'}
+  dangerous_boolean_actions = {:off => 'power off', :reboot => 'power cycle'}
+
+  describe 'safe allocated asset' do
+    safe_allocated_tag = 'safe-allocated-tag'
+    safe_boolean_actions.each do |action, response|
+      it "does #{action}" do
```
roymarantz: perhaps adding "safe" in the it description would be clearer and symmetric with how danger is done.

roymarantz:
This is amazing!!! and looks good to me, but reuse would be improved if:
- the initialize method took and argument, which defaulted to consolr, of the program name
- the start method was implemented something like
  def start options
      process_danger_options options  # process the danger related options and save as instance variables
     # the existing code to get the collins asset
      run asset
  end
  where #run(node) does the dangerous checks and executes the operation if it is allowed.

IMO this would allow people (me) to use this gem for other kinds of safe ipmi hacking.   e.g. nagios check...

support/ruby/consolr/spec/consolr_spec.rb
Primer42: `log` takes an argument (`list` or `clear`). The others take booleans.  I think we should change the way its structured, but this PR was already too big.

Primer42:
@roymarantz - I like all those ideas, but I think I want to do them in future PRs.  This one is already pretty big, and I don't want to change too much at once.

support/ruby/consolr/spec/configs/collins_rspec.yml
```diff
@@ -0,0 +1,4 @@
+host: https://collins.dc.net
```
funzoneq: Best practice is to use example.com: https://tools.ietf.org/html/rfc2606#section-3

funzoneq:
other then the nitpick :+1:

PR #360, opened 2015-08-15
travis improvments

Primer42:
Turns on caching and container builds for travis.

Also, actually fail the tests when the ruby tests fail.

@roymarantz @byxorna @maddalab @defect @gtorre @sushruta

### DIFF
--- a/support/ruby/collins-client/.gitignore
+++ b/support/ruby/collins-client/.gitignore
@@ -5,3 +5,5 @@ rdoc
 .yardoc
 doc
 coverage
+vendor
+.bundle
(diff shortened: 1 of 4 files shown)


support/ruby/collins-client/.gitignore
```diff
@@ -5,3 +5,5 @@ rdoc
 .yardoc
 doc
 coverage
+vendor
+.bundle
```
roymarantz: why don't you want this in git?

byxorna:
LGTM

support/ruby/collins-client/.gitignore
Primer42: We haven't in the past included them.  We could add them, but if we want to then I want to do it in a different PR

PR #362, opened 2015-08-17
net-ping > 1.7.2 requires ruby 1.9.3

Primer42:
Therefore, require 1.7.2 so we can support 1.9.2

@roymarantz @byxorna @maddalab


roymarantz:
:-1:  lets have puppet only install the gem for ruby > 1.9.2 instead

PR #368, opened 2015-09-07
Fix some specs that were not using the right scope

maddalab:
Fix some specs that were not using the right scope, these specs were not actually executing the test. Test count goes from 356 to 377.


maddalab:
`[info] Passed: Total 377, Failed 0, Errors 0, Passed 377`

PR #379, opened 2015-09-28
Implement a firehose for events

maddalab:
Summary:
A firehose implementation using server sent events. Relies
on callback mechanism to implement the firehose. All events
from the firehose return as part of data the asset that is
associated with the event.

All events for assets, asset meta values and ip addresses
are supported.

Every event includes an event name. Supports clustered
installations by making use of hazelcast reliable topic.

The firehose callback processing actor writes events to the
hazelcast reliable topic. We rely on hazelcast to distribute
the event to the cluster.

A single instance of collins installs a listener for the topic,
receives the message and writes it out to all clients of
the firehose.

Prerequisites for using the firehose:

Hazelcast MUST be enabled, does NOT need to be used for cache. This
deployment is applicable when running in stand alone more.

Callbacks MUST be enabled. Callbacks are also required for
search (Solr).

Configuration:

Hazelcast configuration is captured in conf/hazelcast.xml. The
topic can documentation is at [1].

[1] http://docs.hazelcast.org/docs/latest-dev/manual/html/reliabletopicconfiguration.html#reliable-topic-configuration

### DIFF
--- a/app/Global.scala
+++ b/app/Global.scala
@@ -13,6 +13,7 @@ import play.api.mvc.Results
 import collins.callbacks.Callback
 import collins.controllers.ApiResponse
 import collins.db.DB
+import collins.firehose.Firehose
 import collins.hazelcast.HazelcastHelper
 import collins.logging.LoggingHelper
 import collins.metrics.MetricsReporter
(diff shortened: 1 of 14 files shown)


maddalab:
@roymarantz @Primer42 @byxorna @defect

I'm going to do the tests separately.

PR #380, opened 2015-10-02
--pool is a required argument to collins-shell ip_address create

ssgelm:
Updating this adds --pool=POOL to the help text for collins-shell ip_address create which means it is possible to get an ip_address create right on the first try.  Also accurate help text is useful :)

### DIFF
--- a/support/ruby/collins-shell/lib/collins_shell/ip_address.rb
+++ b/support/ruby/collins-shell/lib/collins_shell/ip_address.rb
@@ -101,7 +101,7 @@ module CollinsShell
     method_option :address, :required => true, :type => :string, :desc => 'IP address'
     method_option :gateway, :required => true, :type => :string, :desc => 'IP gateway'
     method_option :netmask, :required => true, :type => :string, :desc => 'IP netmask'
-    method_option :pool, :type => :string, :desc => 'Name of pool'
+    method_option :pool, :required => true, :type => :string, :desc => 'Name of pool'
     def create
       call_collins get_collins_client, "create address" do |client|
         address = client.ipaddress_update! options.tag, nil,


ssgelm:
I tested both with and without a default pool set in the collins config and the ip_address create still failed without a pool.

maddalab:
Thanks for this. I'll dig in more tomorrow morning. However can you provide more description on what the failure was. The [create action](https://github.com/tumblr/collins/blob/master/app/collins/controllers/actions/ipaddress/CreateAction.scala#L34) definitely specifies pool as optional

ssgelm:
Every time I try without the pool I get:
```Error processing request: Specified pool is invalid``
If pool is not required I still think `--pool=POOL` should be listed in the help text in parentheses to indicate that it exists but is optional - I find this confusing every single time I use create (luckily I can usually just use allocate but that's a whole different mess I created for myself :smile:)

**Comment by ssgelm** [comment]
I think the reason that I need to specify a pool is because I don't have defaultPoolName set.  So that would explain the error.  Having said that I really think the help text should indicate that `--pool` is a supported option.

**Comment by maddalab** [comment]
Here is what I found, from a help text stand point
```
➜  ~  collins-shell ip_address
Tasks:
  collins-shell ip_address allocate POOL --tag=TAG                                                 # allocate addresses for an asset in the specified pool
  collins-shell ip_address assets POOL                                                             # find all assets in a pool
  collins-shell ip_address create --address=ADDRESS --gateway=GATEWAY --netmask=NETMASK --tag=TAG  # create a new IP address (Use allocate, not create)
  collins-shell ip_address delete POOL --tag=TAG                                                   # delete addresses for an asset in the specified pool
  collins-shell ip_address delete_all --tag=TAG                                                    # delete all addresses for an asset
  collins-shell ip_address find ADDRESS                                                            # find asset using specified IP address
  collins-shell ip_address help [COMMAND]                                                          # Describe subcommands or one specific subcommand
  collins-shell ip_address pools                                                                   # find all pools that are in use
  collins-shell ip_address update OLD_ADDRESS --tag=TAG                                            # update the IP info for an asset

➜  ~  collins-shell ip_address help create
Usage:
  collins-shell ip_address create --address=ADDRESS --gateway=GATEWAY --netmask=NETMASK --tag=TAG

Options:
  [--config=CONFIG]      # YAML configuration file
  [--debug]              # Debug output
  [--host=HOST]          # Collins host (e.g. http://host:port)
  [--password=PASSWORD]  # Collins password
  [--quiet]              # Be quiet when appropriate
  [--timeout=N]          # Collins client timeout
                         # Default: 30
  [--username=USERNAME]  # Collins username
  --tag=TAG              # Tag for asset
  --address=ADDRESS      # IP address
  --gateway=GATEWAY      # IP gateway
  --netmask=NETMASK      # IP netmask
  [--pool=POOL]          # Name of pool

create a new IP address (Use allocate, not create)
```

Pool is included in the help text, only in detailed help since it is optional.

However I'm not satisfied that you concern has been addressed. I read thru the collins code as well but come up empty. Can you provide more of a description a sequence of steps so that I'm sure we are solving what you want addressed?

ssgelm:
@maddalab thanks so much for looking into it. I'm guessing my issue is that I must not have properly configured a default vlan when testing this. Let me see if I can reproduce and I will let you know.

ssgelm:
@maddalab I think your assessment is correct.  I will close this PR now.

PR #382, opened 2015-10-05
Upgrade activator to 1.3.6 from 1.3.4

maddalab:


roymarantz:
Some description of why you're doing the and what the changes will be would be appreciated.

maddalab:
The PR title is pretty much what the changes will be. I did not check the changelog for activator between 1.3.4 to 1.3.6. I verified that everything builds and runs correctly.

PR #386, opened 2015-12-11
[ipmi] allow templating of asset tag in config

schallert:
It can be useful to have access to an asset's tag in the configuration file for IPMI commands. This enables the asset tag to be accessed as `<tag>` similar to other substitutions. A separate PR will update the docs.


Primer42:
LGTM

byxorna:
LGTM

PR #390, opened 2015-12-30
update dockerfile to jdk8 and cleanup build

byxorna:
@tumblr/systems This dockerfile has gone unloved for a while, so I thought I would simplify and cleanup some. This appears to improve startup times drastically (not writing GC logs to the overlay FS helps :P). JDK bumped to 8, activator to 1.3.7.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,11 +3,9 @@ MAINTAINER Gabe Conradi <gabe@tumblr.com>
 
 RUN apt-get update && apt-get install -y zip unzip && rm -r /var/lib/apt/lists/*
 
-ENV ACTIVATOR_VERSION 1.3.7
-
-# get Play, Collins, build, and deploy it to /opt/collins
 COPY . /build/collins
 RUN cd /build && \
+    export ACTIVATOR_VERSION=1.3.7 && \
     wget -q http://downloads.typesafe.com/typesafe-activator/$ACTIVATOR_VERSION/typesafe-activator-$ACTIVATOR_VERSION-minimal.zip -O /build/activator.zip && \
     unzip -q ./activator.zip && \
     cd collins && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,28 @@
-FROM java:7
+FROM java:8-jdk
 MAINTAINER Gabe Conradi <gabe@tumblr.com>
 
 RUN apt-get update && apt-get install -y zip unzip && rm -r /var/lib/apt/lists/*
 
-RUN useradd -Ur -d /opt/collins collins
-RUN for dir in /build /build/collins /var/log/collins /var/run/collins; do mkdir $dir; chown collins $dir; done
+ENV ACTIVATOR_VERSION 1.3.7
 
-WORKDIR /build
 # get Play, Collins, build, and deploy it to /opt/collins
 COPY . /build/collins
-RUN wget -q http://downloads.typesafe.com/typesafe-activator/1.3.6/typesafe-activator-1.3.6-minimal.zip -O /build/typesafe-activator-1.3.6-minimal.zip && \
-    unzip -q ./typesafe-activator-1.3.6-minimal.zip && \
+RUN cd /build && \
+    wget -q http://downloads.typesafe.com/typesafe-activator/$ACTIVATOR_VERSION/typesafe-activator-$ACTIVATOR_VERSION-minimal.zip -O /build/activator.zip && \
+    unzip -q ./activator.zip && \
     cd collins && \
     java -version 2>&1 && \
-    PLAY_CMD=/build/activator-1.3.6-minimal/activator ./scripts/package.sh && \
+    PLAY_CMD=/build/activator-$ACTIVATOR_VERSION-minimal/activator ./scripts/package.sh && \
     unzip -q /build/collins/target/collins.zip -d /opt/ && \
-    cd / && rm -rf /build && \
-    chown -R collins /opt/collins
-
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,39 +1,28 @@
...
+# expose HTTP, JMX
+EXPOSE 9000 3333
+CMD ["/usr/bin/java","-server","-Dconfig.file=/opt/collins/conf/production.conf","-Dhttp.port=9000","-Dlogger.file=/opt/collins/conf/logger.xml","-Dnetworkaddress.cache.ttl=1","-Dnetworkaddress.cache.negative.ttl=1","-Dcom.sun.management.jmxremote","-Dcom.sun.management.jmxremote.port=3333","-Dcom.sun.management.jmxremote.authenticate=false","-Dcom.sun.management.jmxremote.ssl=false","-XX:MaxPermSize=384m","-XX:+CMSClassUnloadingEnabled","-cp","/opt/collins/lib/*","play.core.server.NettyServer","/opt/collins"]
```
roymarantz: can this extremely long line be split to multiple lines please?

Dockerfile
```diff
@@ -1,39 +1,28 @@
...
-COPY conf/docker/users.conf           /opt/collins/conf/users.conf
-COPY conf/docker/profiles.yaml        /opt/collins/conf/profiles.yaml
-COPY conf/docker/permissions.yaml     /opt/collins/conf/permissions.yaml
-COPY conf/docker/logger.xml           /opt/collins/conf/logger.xml
+WORKDIR /opt/collins
 
-RUN chown -R collins /opt/collins
+# Add in all the default configs we want in this build so collins can run.
+# You probably will want to override these configs in production
+COPY conf/docker conf/
```
roymarantz: do you want to copy or mount this to allow easy updating of the configs (e.g. via puppet) without rebuilding the container?

Dockerfile
byxorna: No, actually. Dockerfile syntax requires it is on one line :(

Dockerfile
byxorna: Configs can be overridden at runtime. This just places the *default* configs in the correct place to enable config-less startup.

roymarantz:
:+1:  assuming it works :-)

Dockerfile
```diff
@@ -1,39 +1,28 @@
-FROM java:7
+FROM java:8-jdk
 MAINTAINER Gabe Conradi <gabe@tumblr.com>
 
 RUN apt-get update && apt-get install -y zip unzip && rm -r /var/lib/apt/lists/*
 
-RUN useradd -Ur -d /opt/collins collins
-RUN for dir in /build /build/collins /var/log/collins /var/run/collins; do mkdir $dir; chown collins $dir; done
+ENV ACTIVATOR_VERSION 1.3.7
```
lpstein: Feels a little off to have this env var sitting around, since it's not used for anything except building the container.  What about moving it to the top of the `RUN` section, rather than having it as a global `ENV`?

lpstein:
LGTM!  Breaking out `ACTIVATOR_VERSION` is a nice touch.

Dockerfile
byxorna: SGTM

Dockerfile
qx-xp: Isn't it the best practice to make the user-configurable storage a `VOLUME` ?

byxorna: For storage, yes. These are just configs though (i.e. the app isnt persisting any state into this directory). The user could totally allocate a volume for `/opt/collins/conf` at runtime if they wanted (but collins would not start up without its production config present)

byxorna:
@lpstein ive updated with your suggestion, thanks!

Dockerfile
qx-xp: Agreed that a set of default configs should be baked into the image.

It would be nice to expose the user-serviceable configs by adding `VOLUME /opt/collins/conf` in the Dockerfile, as described at https://docs.docker.com/engine/articles/dockerfile_best-practices/#volume

Dockerfile
```diff
@@ -1,39 +1,26 @@
...
-RUN useradd -Ur -d /opt/collins collins
-RUN for dir in /build /build/collins /var/log/collins /var/run/collins; do mkdir $dir; chown collins $dir; done
-
-WORKDIR /build
-# get Play, Collins, build, and deploy it to /opt/collins
 COPY . /build/collins
-RUN wget -q http://downloads.typesafe.com/typesafe-activator/1.3.6/typesafe-activator-1.3.6-minimal.zip -O /build/typesafe-activator-1.3.6-minimal.zip && \
-    unzip -q ./typesafe-activator-1.3.6-minimal.zip && \
+RUN cd /build && \
+    export ACTIVATOR_VERSION=1.3.7 && \
```
roymarantz: Does this need to be "export"ed or would a simple shell variable set be enough?  i.e. is it used by any subprocesses or just the single shell that the RUN uses?

Dockerfile
byxorna: That will not work. That will tell docker to allocate a volume at runtime *always*, making our default configs inaccessible (shadowed). This works as it is written, and will allow users to provide their own configs as necessary (via `-v` flags)

Dockerfile
byxorna: Doesnt matter. its only within the context of this `RUN` command, so nothing leaks out into the built image.

Dockerfile
qx-xp: I thought it would copy the files into the volume.  But since it doesn't work like that, working code is good. :+1:

PR #391, opened 2016-01-04
Documentation update

### LINKED ISSUE #394: Multiauth documentation is missing
Looks like the docs don't have any details about multiauth. Could be useful dropping some example configs into the documentation

defect:
Adding some documentation about mixed authentication and the firehose.

Fixes #394 

@tumblr/systems

### DIFF
--- a/_data/api.yml
+++ b/_data/api.yml
@@ -2079,11 +2079,14 @@ sections:
             event: asset_update
             data: {"name":"asset_update","tag":"tumblrtag30","category":"Asset","data":{"ASSET":{"ID":1,"TAG":"tumblrtag30","STATE":{"ID":3,"STATUS":null,"NAME":"RUNNING","LABEL":"Running","DESCRIPTION":"A service in this state is operational."},"STATUS":"Allocated","TYPE":"SERVER_NODE","CREATED":"2013-08-28T12:51:00","UPDATED":"2015-12-05T19:53:27","DELETED":null}}}
 
-            event: asset_update
-            data: {"name":"asset_update","tag":"tumblrtag31","category":"Asset","data":{"ASSET":{"ID":2,"TAG":"tumblrtag31","STATE":{"ID":3,"STATUS":null,"NAME":"RUNNING","LABEL":"Running","DESCRIPTION":"A service in this state is operational."},"STATUS":"Allocated","TYPE":"SERVER_NODE","CREATED":"2013-08-28T12:51:00","UPDATED":"2015-12-05T19:53:27","DELETED":null}}}
+            event: asset_create
+            data: {"name":"asset_create","tag":"tumblrtag31","category":"Asset","data":{"ASSET":{"ID":2,"TAG":"tumblrtag31","STATE":null,"STATUS":"Incomplete","TYPE":"SERVER_NODE","CREATED":"2016-01-13T11:30:37","UPDATED":null,"DELETED":null}}}
 
-            event: asset_update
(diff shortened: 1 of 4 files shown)


_includes/configuration/authentication.html
```diff
@@ -201,3 +201,35 @@
...
+<h3>Mixed Authentication</h3>
+
+<p>
+Mixed authentication allows collins to use multiple authentication backends
+simultaneously. An example use-case could be a scenario where real users live in
+LDAP but system users (for automation etc.) are defined using file
+authentication.
+</p>
+<p>
+To use mixed authentication, simply specify a list of authentication backends as
```
roymarantz: Does the order of the list set the search order for the authentication backends or are they searched in parallel?  What happens if the same name is in multiple backends, i.e. which one wins?

_data/api.yml
```diff
@@ -2062,6 +2062,29 @@ sections:
...
+          curl: curl --basic -u blake:admin:first http://localhost:9000/api/firehose
+        responses:
+          SSE: |
+            event: asset_update
+            data: {"name":"asset_update","tag":"tumblrtag30","category":"Asset","data":{"ASSET":{"ID":1,"TAG":"tumblrtag30","STATE":{"ID":3,"STATUS":null,"NAME":"RUNNING","LABEL":"Running","DESCRIPTION":"A service in this state is operational."},"STATUS":"Allocated","TYPE":"SERVER_NODE","CREATED":"2013-08-28T12:51:00","UPDATED":"2015-12-05T19:53:27","DELETED":null}}}
+
+            event: asset_update
+            data: {"name":"asset_update","tag":"tumblrtag31","category":"Asset","data":{"ASSET":{"ID":2,"TAG":"tumblrtag31","STATE":{"ID":3,"STATUS":null,"NAME":"RUNNING","LABEL":"Running","DESCRIPTION":"A service in this state is operational."},"STATUS":"Allocated","TYPE":"SERVER_NODE","CREATED":"2013-08-28T12:51:00","UPDATED":"2015-12-05T19:53:27","DELETED":null}}}
+
+            event: asset_update
```
Primer42: Could we show examples of each of the event types?  Even if the `data` field is identical, that would be a more useful example.
https://github.com/tumblr/collins/blob/abecbf07546d5279a5f4aef36b70b299800cccd3/app/collins/firehose/Firehose.scala#L31-L37

defect: Yeah, i actually didn't realize we had support for more things than `asset_update`s since that's the vast majority of events. As soon as I get the firehose running in my collins container i'll capture some other events too and add them.

_includes/configuration/authentication.html
defect: Added a sentence clarifying the order and which backend "wins".

_includes/configuration/authentication.html
```diff
@@ -201,3 +201,36 @@
...
+the type. The mixed authentication provider will try each of the methods listed,
```
Primer42: Just to be explicit, I would tweak this to say 
> The mixed authentication provider will try each of the methods **in the order listed** and use the first one that returns a successful match.

_data/api.yml
defect: Added `asset_create`, `asset_delete` and `ipAddresses_create` examples.

_includes/configuration/authentication.html
Primer42: Actually, you need to provide a list of the authentication methods described, as well as their configs as outlined in the other sections.  Can we clarify that a bit?

defect: Something like this maybe? 30f931

Primer42:
LGTM

PR #392, opened 2016-01-05
Refactor/scalaish

baloo:
Small editions to make code looks a bit more scalaish.

Those are untested (not having softlayer account in hands)

### DIFF
--- a/app/collins/softlayer/SoftLayer.scala
+++ b/app/collins/softlayer/SoftLayer.scala
@@ -51,18 +51,16 @@ trait SoftLayer {
 
 object SoftLayer extends SoftLayer {
 
+  private val softLayerIdPattern = "^sl-(\\d+)$".r
+
   // start plugin API
-  override def isSoftLayerAsset(asset: Asset): Boolean = {
-    asset.tag.startsWith("sl-")
+  override def isSoftLayerAsset(asset: Asset): Boolean = asset.tag match {
+    case softLayerIdPattern(_*) => true
+    case _ => false
   }
-  override def softLayerId(asset: Asset): Option[Long] = isSoftLayerAsset(asset) match {
-    case true => try {
-      Some(asset.tag.split("-", 2).last.toLong)
-    } catch {
-      case _: Throwable => None
-    }
-    case false => None
-  }
+
+
+  override def softLayerId(asset: Asset): Option[Long] = (softLayerIdPattern findFirstIn asset.tag).map(_.toLong)
 
   private[this] val TicketExtractor = "^.* ([0-9]+).*$".r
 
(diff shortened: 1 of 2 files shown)


app/collins/softlayer/SoftLayer.scala
```diff
@@ -94,10 +92,13 @@ object SoftLayer extends SoftLayer {
...
     val url = softLayerUrl("/SoftLayer_Hardware_Server/%d/sparePool.json".format(id))
-    val query = JsObject(Seq("parameters" -> JsArray(List(JsString("activate")))))
-    val queryString = Json.stringify(query)
+    val query = Json.obj(
+      "parameters" -> Json.arr(
+         "activate"
+      )
+    )
 
-    val wsUrl = WS.url(url.toString).withHeaders("Content-Type" -> "application/json", "Content-Length" -> queryString.length.toString)
```
byxorna: Where are you setting these headers when making the request?

byxorna:
This LGTM. We don't use this plugin either, so we cant really test this.

app/collins/softlayer/SoftLayer.scala
baloo: Play will make an in-memory buffer and pass it to ning (https://github.com/playframework/playframework/blob/master/framework/src/play-ws/src/main/scala/play/api/libs/ws/WS.scala#L398)

Ning will get the Content-Length part (because http/1.1 says it needs a content-length if not streamed (which is not the case here)). 

The Content-Type is set by play because you pass in a JsValue type and Content-Type will automatically resolve to 'application/json' **compile-time**. 
Implicit / call-chain is:
 * https://github.com/playframework/playframework/blob/master/framework/src/play-ws/src/main/scala/play/api/libs/ws/WS.scala#L402
 * https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/http/Writeable.scala#L94
 * https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/http/ContentTypeOf.scala#L51
 * https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/http/StandardValues.scala#L29
 * https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/http/StandardValues.scala#L110

Related documentation: https://www.playframework.com/documentation/2.3.x/ScalaWS#Submitting-JSON-data

baloo:
bumping commit to get the tests passing (works on my machine, looks like a timeout when running on travis).

baloo:
Ok, I can't figure out why tests are failing, works on my machine. Last commit tries to get higher timeouts. If you have any idea, please raise your voice.

maddalab:
the tests are a timing issue with solr indexing due to resource allocation on travis. Sometimes have to just be kicked off once more. I did that they succeed now. ty for this.

PR #396, opened 2016-01-19
Include solr query string in cache key

defect:
When doing queries across many collins instances (using multi-collins) we cache the response from the remote instances for 30 seconds. The problem is that we do not include the CQL query in the key which can result in two different queries returning the same thing.

This PR changes the behavior by adding the solr query string to the cache key, which should eliminate the problem. I'm "unprotecting" `traverseQueryString()` but I think that should be fine as the method doesn't modify or remove anything, it just traverses the solr expression and builds the query string.

@tumblr/systems rfr

### DIFF
--- a/app/collins/models/AssetFinder.scala
+++ b/app/collins/models/AssetFinder.scala
@@ -54,7 +54,7 @@ case class AssetFinder(
       updatedAfter.map(t => "updatedAfter" -> Formatter.dateFormat(t)) ::
       updatedBefore.map(t => "updatedBefore" -> Formatter.dateFormat(t)) ::
       state.map(s => "state" -> s.name) ::
-      query.map { q => "query" -> "UHOH!!!!" } :: //FIXME: need toCQL traversal
+      query.map(q => "query" -> q.traverseQueryString) ::
       Nil)
     items.flatten
   }
--- a/app/collins/models/RemoteAssetFinder.scala
+++ b/app/collins/models/RemoteAssetFinder.scala
@@ -276,7 +276,7 @@ object RemoteAssetFinder {
   /**
    */
   def apply(clients: Seq[RemoteAssetClient], pageParams: PageParams, searchParams: AssetSearchParameters): (Seq[AssetView], Long) = {
-    val key = searchParams.paginationKey + clients.map { _.tag }.mkString("_")
+    val key = searchParams.paginationKey.getOrElse("") + clients.map { _.tag }.mkString("_")
     val stream = Cache.getAs[RemoteAssetStream](key).getOrElse(new RemoteAssetStream(clients, searchParams))
     val results = stream.slice(pageParams.page * pageParams.size, (pageParams.page + 1) * (pageParams.size))
     Cache.set(key, stream, 30)
--- a/app/collins/models/RemoteAssetFinder.scala
+++ b/app/collins/models/RemoteAssetFinder.scala
@@ -26,6 +26,7 @@ import collins.solr.SolrOrOp
 import collins.solr.StringValueFormat
 import collins.util.AttributeResolver.ResultTuple
 import collins.util.RemoteCollinsHost
+import collins.util.config.MultiCollinsConfig
 
 /**
  * Just a combination of everything needed to do a search.  Probably should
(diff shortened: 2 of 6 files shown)


app/collins/models/AssetFinder.scala
```diff
@@ -54,7 +54,7 @@ case class AssetFinder(
       updatedAfter.map(t => "updatedAfter" -> Formatter.dateFormat(t)) ::
       updatedBefore.map(t => "updatedBefore" -> Formatter.dateFormat(t)) ::
       state.map(s => "state" -> s.name) ::
-      query.map { q => "query" -> "UHOH!!!!" } :: //FIXME: need toCQL traversal
```
tbchrist: This is magical.

bobpattersonjr: this might be my favorite bug

app/collins/models/RemoteAssetFinder.scala
```diff
@@ -276,7 +276,7 @@ object RemoteAssetFinder {
   /**
    */
   def apply(clients: Seq[RemoteAssetClient], pageParams: PageParams, searchParams: AssetSearchParameters): (Seq[AssetView], Long) = {
-    val key = searchParams.paginationKey + clients.map { _.tag }.mkString("_")
+    val key = searchParams.paginationKey.getOrElse("") + clients.map { _.tag }.mkString("_")
     val stream = Cache.getAs[RemoteAssetStream](key).getOrElse(new RemoteAssetStream(clients, searchParams))
     val results = stream.slice(pageParams.page * pageParams.size, (pageParams.page + 1) * (pageParams.size))
     Cache.set(key, stream, 30)
```
bobpattersonjr: should this stay at 30?

defect: I don't really see a reason to change it right now, it's been a while since we actually used the multi-collins functionality in production so i suggest keeping it at 30 to see how it works out. I'll open a ticket to move it to a configuration parameter though.

app/collins/solr/CollinsQueryLanguage.scala
```diff
@@ -27,7 +27,7 @@ case class CQLQuery(select: SolrDocType, where: SolrExpression) {
  * itself into a Solr Query
  */
 sealed trait SolrQueryComponent {
-  protected[solr] def traverseQueryString(): String = traverseQueryString(true)
```
maddalab: Just a quick skim not really a through check, `toplevel` param is unused.

Also, caching has plenty of tests and this seems like one that is very do-able, ymmv

app/collins/models/RemoteAssetFinder.scala
byxorna: Ideally this would be configurable, but meh.

byxorna:
:nail_care: game on fleek @defect!

app/collins/models/RemoteAssetFinder.scala
defect: Yeah, i just pushed a commit to add it to config. It still defaults to 30 seconds but is now configurable as `multicollins.queryCacheTimeout`.

conf/reference/multicollins_reference.conf
```diff
@@ -3,4 +3,5 @@ multicollins {
   instanceAssetType = DATA_CENTER
   locationAttribute = LOCATION
   thisInstance = NONE
+  queryTimeout = 30
```
byxorna: Why is this value `queryTimeout` but the config class is referencing `queryCacheTimeout`?

defect: Because i'm sloppy :) Fix incoming.

app/collins/solr/CollinsQueryLanguage.scala
defect: It's used in one spot: https://github.com/tumblr/collins/blob/master/app/collins/solr/CollinsQueryLanguage.scala#L140

Expressions that have multiple children are not wrapped in parentheses if they are the top level expression. That being said, it shouldn't change anything to actually wrap them, but i'll open another PR to simplify things by removing the `toplevel` parameter in order to keep this one focused on fixing the query caching.

defect: https://github.com/tumblr/collins/pull/397

Primer42:
LGTM - @defect please also be sure to make a PR to update the multicollins parameters docs. https://github.com/tumblr/collins/blob/gh-pages/_includes/configuration/multicollins.html#L23-51

defect:
@Primer42 #399 opened

roymarantz:
Do documentation changes go in a different PR?  How about updated tests?
The code LGTM, but this is not my favorite language :-)

defect:
Yes, documentation goes in a different PR (#399) since those are pulled in to another branch (gh-pages). I also added a test to make sure the cql query gets included.

schallert:
LGTM. What a find :shipit:

roymarantz:
:+1:

PR #397, opened 2016-01-20
Remove toplevel parameter from solr

defect:
The `toplevel` parameter of the `traverseQueryString()` method is only used once, and in that case can be removed without changing the structure of the query. So in the interest of simplifying things slightly i propose to simply remove it.

@tumblr/systems @maddalab


Primer42:
LGTM

byxorna:
@defect how is this tested?

defect:
@byxorna I've been running this locally for smoke testing and haven't run in to any problems yet. Also, the test suite is happy :)

byxorna:
@defect awesome, lgtm! :cake:

PR #400, opened 2016-01-21
Revert "Include solr query string in cache key"

defect:
Reverts tumblr/collins#396

An error is thrown when sending the query to the remote instances. I suspect the problem lies in URL encoding but I haven't tracked it down quite yet. I'll open a new PR once i do.

@tumblr/collins


schallert:
:+1:

Primer42:
LGTM

PR #401, opened 2016-01-25
Make remote query cache timeout configurable

defect:
Fixing the problems with multi-collins queries as described in #396 (and reverted in #400) is taking a bit longer than i had hoped. So, in the meantime, this PR re-introduces the config parameter for the remote query cache timeout from the previous PR.

@tumblr/collins

### DIFF
--- a/app/collins/models/RemoteAssetFinder.scala
+++ b/app/collins/models/RemoteAssetFinder.scala
@@ -26,6 +26,7 @@ import collins.solr.SolrOrOp
 import collins.solr.StringValueFormat
 import collins.util.AttributeResolver.ResultTuple
 import collins.util.RemoteCollinsHost
+import collins.util.config.MultiCollinsConfig
 
 /**
  * Just a combination of everything needed to do a search.  Probably should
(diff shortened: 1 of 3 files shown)


app/collins/models/RemoteAssetFinder.scala
```diff
@@ -279,7 +280,8 @@ object RemoteAssetFinder {
     val key = searchParams.paginationKey + clients.map { _.tag }.mkString("_")
     val stream = Cache.getAs[RemoteAssetStream](key).getOrElse(new RemoteAssetStream(clients, searchParams))
     val results = stream.slice(pageParams.page * pageParams.size, (pageParams.page + 1) * (pageParams.size))
-    Cache.set(key, stream, 30)
+    val timeout = MultiCollinsConfig.queryCacheTimeout
```
roymarantz: should this do something different if multi-collins is not enabled?

byxorna:
LGTM

app/collins/models/RemoteAssetFinder.scala
defect: This code path will never be hit if multi-collins isn't enabled. In order to do a multi-collins query `MultiCollinsConfig.enabled` must be true and the `remoteLookup` parameter specified: https://github.com/tumblr/collins/blob/master/app/collins/controllers/actions/asset/FindAction.scala#L45

I can't really think of a case where we'd want to cache results from different instances for different amounts of time. I think for the scope of this PR i'd like to leave that out. We'd probably have to store the timeout in the datacenter asset which makes things a bit more complicated.

roymarantz: latency and timeliness differences between sites could be a reason, but we are far enough away from that that, IMO, you should ignore this, especially if it is hard to do

roymarantz:
:+1:

PR #402, opened 2016-01-27
add volume for solr cores

byxorna:
Solr cores should not be written to the rootfs. The are considered state, so they should have a volume allocated. cc @Primer42 @defect @roymarantz @schallert


Primer42:
LGTM

roymarantz:
:+1:

schallert:
LGTM

PR #405, opened 2016-01-28
format provisioning errors less shittily

byxorna:
Provisioning errors are hard to read right now. Change that to format preformatted code with like... `<pre><code>...</code></pre>`. @Primer42 @defect @schallert @roymarantz @maddalab

### DIFF
--- a/public/javascripts/main.js
+++ b/public/javascripts/main.js
@@ -235,7 +235,7 @@ $(document).ready(function() {
           var message = response.data.message;
           html += '<div style="font-family: monaco">' + message + '</div>';
           if (details && typeof(details) === "object" && details.message) {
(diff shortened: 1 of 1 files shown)


Primer42:
LGTM

roymarantz:
:+1:  for the code, but does the "caller" have any trouble with this?  it is likely to change the "shape" of the  displayed message.  e.g if the message used to be displayed as a single line and the formating shapes it into multi-line...

byxorna:
@roymarantz this is web view only. "clients" are humans clicking buttons. This does not impact responses given over API at all

roymarantz:
that is what I was hoping for.  thaks

byxorna:
@Primer42 merge?

PR #407, opened 2016-02-01
Fix multi-collins queries

defect:
This is a follow-up to #396 and tries to fix a couple of issues with multi-collins queries.

**1. Send CQL query to remote instances** (848a11a)

This is what #396 tried to fix but didn't since I actually sent the query in Solr format instead of CQL to the remote instances. To fix this, I've changed the way we handle CQL queries by storing them as a `String` instead of a `SolrExpression`, and only parse the query to solr syntax when we're actually planning to execute it. This means that we can pass the raw CQL query string to the remote instances.

**2. Fix parsing of remote responses where state is null** (04571fd)

Previously, if an asset a remote instance returned didn't have a state, the JSON parser would fail on the null value.

**3. Fix filtering of the local instance** (9025bcd)

When doing a multi-collins query we try to filter out the local instance (the one we actually send the query to) based on the `multicollins.thisInstance` config variable. However, we're comparing a `String` with an `Option[String]` which fails so we'll actually get the results from the local instance twice, once for the multi-collins query and once for the local.

@tumblr/collins @tbchrist @bobpattersonjr

CC'ing you too @cburroughs, since it would be interesting to hear feedback based on your multi-collins use cases.


app/collins/models/Asset.scala
```diff
@@ -266,7 +266,7 @@ object Asset extends Schema with AnormAdapter[Asset] with AssetKeys {
     val findLocations = find(PageParams(0, 50, "ASC", "tag"), AttributeResolver.emptyResultTuple, instanceFinder)
       .items
       .collect { case a: Asset => a }
-      .filter { _.tag != MultiCollinsConfig.thisInstance }
+      .filter(asset => { !MultiCollinsConfig.thisInstance.contains(asset.tag) })
```
roymarantz: t would be nice to add a comment explaining this.  On the surface it looks like you are removing any assets that exist in the current instance, instead of assets returned from the direct (vs multi-collins) query.  [unless I'm misinterpreting things...]

app/collins/models/conversions.scala
```diff
@@ -136,7 +137,10 @@ object conversions {
       (json \ "UPDATED").asOpt[Timestamp],
       (json \ "DELETED").asOpt[Timestamp],
       (json \ "ID").as[Long],
-      (json \ "STATE").asOpt[State].map(_.id).getOrElse(0)))
+      (json \ "STATE") match {
+        case JsNull => State.ANY_STATUS
```
roymarantz: I don't understand how state can be null since it should be defined once the asset is created unless Unknown translates to null???

roymarantz:
:+1:  if this works ,since it is hard to argue with success :-) , even with my comments

app/collins/models/Asset.scala
defect: This is just for finding the datacenter assets. Before actually executing the query the user sent, we have to figure out which remote collins instances are around that we need to send the query to. 

What we actually do here is query for assets of type `MultiCollinsConfig.instanceAssetType` which is the asset type that represents instances in other datacenters. This filter will remove one asset, and that is the local datacenter.

app/collins/models/conversions.scala
defect: I think you might be confusing Status (`New`, `Incomplete`, `Allocated` etc.) and State (`RUNNING`, `MAINT_NOOP` etc). A newly created asset will have a status, but not a state.

roymarantz: If you say so.  I would think State and Status should always be well defined, but reality and (my) theory often disagree.

app/collins/models/Asset.scala
roymarantz: If you think the code has an "obvious" function, then OK, else a comment would be nice.

defect: I agree, it's not very obvious. Comment added.

app/collins/models/Asset.scala
```diff
@@ -266,7 +266,8 @@ object Asset extends Schema with AnormAdapter[Asset] with AssetKeys {
     val findLocations = find(PageParams(0, 50, "ASC", "tag"), AttributeResolver.emptyResultTuple, instanceFinder)
       .items
       .collect { case a: Asset => a }
-      .filter { _.tag != MultiCollinsConfig.thisInstance }
+      // filter out the local instance so that we don't perform both a local and a remote query to it
+      .filter(asset => { !MultiCollinsConfig.thisInstance.contains(asset.tag) })
```
Primer42: So, this is a really perverse case, but I think this would mess up if you had a datacenter instance name that was a substring of another datacenter name.  For example, if your datacenters were `DC` and `DC1`.  I think you may want to do an equality match here, maybe even make it case insensitive.

app/collins/models/RemoteAssetFinder.scala
```diff
@@ -48,15 +48,15 @@ case class AssetSearchParameters(
   def toSeq: Seq[(String, String)] = {
     val q1: Seq[(String, String)] = (
       params._1.map { case (enum, value) => (enum.toString, value) } ++
-      params._2.map { case (assetMeta, value) => ("attribute" -> "%s;%s".format(assetMeta.name, URLEncoder.encode(value, "UTF-8"))) } ++
-      params._3.map { i => ("attribute" -> ("ip_address;" + URLEncoder.encode(i, "UTF-8"))) }) ++ afinder.toSeq :+ ("details" -> (if (details) "true" else "false"))
+      params._2.map { case (assetMeta, value) => ("attribute" -> "%s;%s".format(assetMeta.name, value)) } ++
```
Primer42: Can you explain what is going on here? Is it safe to not URL encode the strings we're sending across over HTTP?

Primer42: Nevermind, you URLEncoded everything later. I didn't read far enough.

app/collins/models/Asset.scala
defect: This isn't a substring match but a function on the `Option` type. It does an equality test on the string within the `Option[String]` that is `MultiCollinsConfig.thisInstance` (or returns false if the option is `None`). It is not a substring match so I think this should be fine.

```
scala> Some("dc1").contains("dc1")
res1: Boolean = true

scala> Some("dc1").contains("dc")
res0: Boolean = false

scala> None.contains("dc1")
res2: Boolean = false
```

app/collins/models/RemoteAssetFinder.scala
defect: Yep, the URL encoding now happens in `toQueryString` which, you know, actually produces a query string :)

app/collins/models/Asset.scala
Primer42: Ahhh OK awesome

Primer42:
LGTM

PR #408, opened 2016-02-03
Collins-shell fix dependencies

Primer42:
A few changes here:
- Remove `jewler` and `rubygems-update` dependencies
- Remove weird backwards compatible jewler code from gemspec file 
- Use gemspec in Gemfile rather than declaring all dependencies twice

@tumblr/collins


support/ruby/collins-shell/collins_shell.gemspec
```diff
@@ -1,95 +1,70 @@
...
+    'lib/collins_shell/util.rb',
+    'lib/collins_shell/util/asset_printer.rb',
+    'lib/collins_shell/util/asset_stache.rb',
+    'lib/collins_shell/util/log_printer.rb',
+    'lib/collins_shell/util/printer_util.rb'
   ]
-  s.homepage = "https://github.com/tumblr/collins/tree/master/support/ruby/collins-shell"
-  s.licenses = ["APL 2.0"]
-  s.require_paths = ["lib"]
-  s.rubygems_version = "1.8.23"
```
Primer42: Docs say that this should never be set by hand, so I deleted it
http://guides.rubygems.org/specification-reference/#rubygems_version

komapa:
:bullettrain_front: LGTM

roymarantz:
:+1:

Primer42:
The scala tests failed, but all the ruby tests passed, so this is safe to go out.

PR #412, opened 2016-02-15
Consolr updates

defect:
In order to support more tools than ipmitool in consolr, I'd like to modularize it a bit. As you can see here, I've moved all the ipmitool-related code in to a class called `Consolr::Runners::Ipmitool`. The command loads one or more runners dynamically from the config, and chooses the first one that reports it can run on a given node (meaning the runner's `can_run?` method returns true). 

An example use case could be an environment with mixed Openstack VMs and bare metal servers. An openstack runner would then handle all the VM assets and the ipmitool runner all the bare metal assets. This way, you can easily customize consolr to new tools by writing a `Consolr::Runners::Yourtool` class and then installing it in whichever way you prefer.

In order to maintain backwards compatibility I've made the ipmitool runner the default if no runners are specified in the config.

Finally, I'm adding an option (`-S`, or `--status`) which just returns the power status of the assets because it seemed like something that would be useful :)

@tumblr/collins

### DIFF
--- a/support/ruby/consolr/lib/consolr/runners/ipmitool.rb
+++ b/support/ruby/consolr/lib/consolr/runners/ipmitool.rb
@@ -55,8 +55,16 @@ module Consolr
         cmd 'power off', node
       end
 
+      def soft_off node
+        cmd 'power soft', node
+      end
+
       def reboot node
         cmd 'power cycle', node
+      end
+
+      def soft_reboot node
+        cmd 'power reset', node
       end
 
       def status node
--- a/support/ruby/consolr/lib/consolr/runners/ipmitool.rb
+++ b/support/ruby/consolr/lib/consolr/runners/ipmitool.rb
@@ -4,10 +4,18 @@ module Consolr
   module Runners
     class Ipmitool < Runner
       def initialize config
-				@ipmitool = config['executable']
+        @ipmitool = if config.empty?
+                      '/usr/bin/ipmitool'
+                    else
+                      config
+                    end
       end
 
-			def verify? node
+      def can_run? node
+        not (node.ipmi.address.empty? or node.ipmi.username.empty? or node.ipmi.password.empty?)
+      end
+
+			def verify node
       	Net::Ping::External.new(node.ipmi.address).ping?
 			end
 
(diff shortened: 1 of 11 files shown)


byxorna:
LGTM

support/ruby/consolr/lib/consolr/runners/ipmitool.rb
```diff
@@ -0,0 +1,70 @@
...
+                    end
+      end
+
+      def can_run? node
+        not (node.ipmi.address.empty? or node.ipmi.username.empty? or node.ipmi.password.empty?)
+      end
+
+			def verify node
+      	Net::Ping::External.new(node.ipmi.address).ping?
+			end
```
roymarantz: [picky] whitespace

support/ruby/consolr/lib/consolr/runners/ipmitool.rb
```diff
@@ -0,0 +1,70 @@
...
+      def initialize config
+        @ipmitool = if config.empty?
+                      '/usr/bin/ipmitool'
+                    else
+                      config
+                    end
+      end
+
+      def can_run? node
+        not (node.ipmi.address.empty? or node.ipmi.username.empty? or node.ipmi.password.empty?)
```
roymarantz: do you want a rescue false incase node.ipmi (or some part) is nil?

support/ruby/consolr/lib/consolr/runners/runner.rb
```diff
@@ -0,0 +1,56 @@
...
+module Consolr
+  module Runners
+    class Runner
+      def can_run? node
+        raise 'can_run? is not implemented for this runner'
+      end
+
+			def verify node
+        raise 'verify is not implemented for this runner'
+			end
```
roymarantz: [picky] indentation

roymarantz:
IMO there should be some way to force using a particular runner.
Do you want to add missing ipmitool functions in the PR (e.g. soft reboot vs hard reboot) or is that too much out of scope.
I like the approach though.

gtorre:
Very nice. :+1:

defect:
Added support for `--soft-reboot` and `--soft-off`.

Primer42:
This LGTM
But, I think we should add a barebones example gem that would show people how to add their own, internal runners with custom logic.

Primer42:
Can you also add unit tests before merging?

PR #419, opened 2016-02-24
Dockerfile: use the JRE image, not JDK

unclejack:
This PR switches the Dockerfile image to the JRE image from the JDK. This reduces the size of the final image by 331 MB because it doesn't have to pull some of those big layers from the JDK image.

The number of layers has also been reduced through this change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,14 @@
-FROM java:8-jdk
+FROM java:8-jre
 MAINTAINER Gabe Conradi <gabe@tumblr.com>
-
-RUN apt-get update && apt-get install -y zip unzip && rm -r /var/lib/apt/lists/*
 
 # Solr cores should be stored in a volume, so we arent writing stuff to our rootfs
 VOLUME /opt/collins/conf/solr/cores/collins/data
 
 COPY . /build/collins
-RUN cd /build && \
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y openjdk-8-jdk zip unzip && \
+    rm -r /var/lib/apt/lists/* && \
+    cd /build && \
     export ACTIVATOR_VERSION=1.3.7 && \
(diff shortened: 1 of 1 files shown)


byxorna:
@unclejack nice! Curious why this actually results in a smaller size image; the only difference I see between the JRE and JDK dockerfiles is that jdk installs `openjdk-8-jdk="$JAVA_DEBIAN_VERSION"`, and the jre installs `openjdk-8-jre-headless="$JAVA_DEBIAN_VERSION"`, and the JDK image comes from `buildpack-deps:jessie-scm` instead of `buildpack-deps:jessie-curl`. Wont argue with success though!

LGTM!

unclejack:
@byxorna There are a few reasons behind this reduction in size. First off, the fact that I'm cleaning up the temporary folders in the root directory saves about ~389 megabytes from the final image. This was something I've discovered after mentioning that original 331 MB reduction in size as those got even bigger since the last image was pushed to the Hub (that's what I was comparing to when reporting my results).
Another difference in size comes from the fact that the JDK image is also installing some other packages such as mercurial, git, openssh-client, bzr and subversion.

byxorna:
@unclejack huge :+1:. I didnt think that the ivy cache would persist somewhere permanent, instead of something like `/var/tmp`. @Primer42 merge this in!

Primer42:
Looks awesome - thanks @unclejack

unclejack:
@byxorna @Primer42 I've pushed an update. The resulting image is just 422 MB now. The layer which contains the result of the build is now just 110.5 MB; this layer was 368 MB before.

byxorna:
@unclejack impressive. @Primer42 ship this before @unclejack ports collins to a unikernel!

Primer42:
Hi @unclejack - would you mind signing our CLA before we merge this?
https://github.com/tumblr/collins/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla
I know it isn't a large change, just would rather be safe than sorry with legal-related things like this.
Thanks!

PR #423, opened 2016-03-01
Add support for sensor reading

defect:
`ipmitool sensor list` spits out a long list of sensors, their value and their configured thresholds. `sdr elist all` gives a similar list, but without some of the information so i'd like to add support to grab the full sensors output.

@tumblr/collins


roymarantz:
:+1:  and I guess this answers my comment from the other PR :-)

byxorna:
:+1:

PR #424, opened 2016-03-03
add a flag for consolor to print SOL info

sushruta:
Adding SOL info for the host we wish to connect to. This can give us session info in case we need to need to dig deeper into session count and handles.

usage -- `consolr --tag <asset_tag> --sol-info`

example output

```
session handle                : 122
slot count                    : 36
active sessions               : 1
user id                       : 2
privilege level               : ADMINISTRATOR
session type                  : IPMIv2/RMCP+
channel number                : 0x01
console ip                    : 10.11.12.13
console mac                   : 4c:26:24:4e:da:21
console port                  : 29854

```

### DIFF
--- a/support/ruby/consolr/bin/consolr
+++ b/support/ruby/consolr/bin/consolr
@@ -12,6 +12,7 @@ opt_parser = OptionParser.new do |opt|
   opt.on('-c', '--console',         'console into node via SOL')               { options[:console] = true }
   opt.on('-d', '--dangerous',       'list dangerous stuff')                    { options[:dangerous] = true }
   opt.on('-f', '--force',           'force run dangerous actions')             { options[:force] = true }
+  opt.on('-g', '--get-sol-info',    'get Serial Over LAN (SOL) info')          { options[:get_sol_info] = true }
   opt.on('-H', '--hostname ASSET',  'asset hostname')                          { |hostname| options[:hostname] = hostname }
   opt.on('-i', '--identify',        'turn on chassis UID')                     { options[:identify] = true }
   opt.on('-k', '--kick',            'kick if someone is hogging the console')  { options[:kick] = true }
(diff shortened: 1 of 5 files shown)


roymarantz:
:+1:  if it works :-)

defect:
lgtm

support/ruby/consolr/bin/consolr
```diff
@@ -17,6 +17,7 @@ opt_parser = OptionParser.new do |opt|
   opt.on('-k', '--kick',            'kick if someone is hogging the console')  { options[:kick] = true }
   opt.on('-l', '--log LOG',         'System Event Log (SEL) [list|clear]')     { |log| options[:log] = log }
   opt.on('-o', '--on',              'turn on node')                            { options[:on] = true }
+  opt.on('-p', '--sol-info',        'get Serial Over LAN (SOL) info')          { options[:get_sol_info] = true }
```
Primer42: How did you decide on p?

Primer42:
Please remember to bump the version - https://github.com/tumblr/collins/blob/master/support/ruby/consolr/lib/consolr/version.rb

support/ruby/consolr/bin/consolr
defect: Tiny nitpick: i think that in all other cases we're using the long parameter name as the symbol name in the options hash. For instance, `--on` => `options[:on]`, `--kick` => `options[:kick]` etc. Might want to keep with that convention.

Not going to block you on that though. My LGTM still stands.

sushruta: I changed it to `-g` for "get-sol-info"

sushruta:
do things look fine now @Primer42?

Primer42:
Please rebase and squash before merging.
LGTM

PR #427, opened 2016-03-17
disable PerfData inside container

byxorna:
In containers, /tmp isnt on tmpfs by default, so the default `-XX:+UsePerfData` jvm option really sucks. It creates a huge amount of IO on the rootfs because the `/tmp/hsperfdata_root/1` file keeps changing. (you know... for performance).

Lets nuke that shit. @Primer42 @schallert @roymarantz @maddalab @mschenck @defect

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,5 +25,5 @@ COPY conf/docker conf/
 
 # expose HTTP, JMX
 EXPOSE 9000 3333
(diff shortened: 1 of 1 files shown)


Primer42:
LGTM

byxorna:
@Primer42 bump

Issue #422, opened 2016-03-01
gen_passwords.sh is not generating correct hashes


MrRagnar:
I have generated a user and password-hash and put them in the users.yml file but I keep getting
"**Unable to login Invalid username or password**" when I try to login.
Logging in as **blake** with the password **admin:first** works just fine.
To test, I generated the hash for that password using htpasswd:
```
# echo admin:first | htpasswd -i -s -n test
test:{SHA}0sJmRmPZiSG3mjLYl0Zcd1M1W7I=
```
This differs from the hash in the default **users.yml** file:
```
# cat users.conf
blake:{SHA}Xziwl2ZuU/FOR4XXbjOfN7MMRNw=:engineering,infra,ops
```

Am I missing something or is the **gen_passwords.sh** broken? (missing a function to add a salt maybe?)

Primer42:
Hi @MrRagnar 
Sorry for the late response.
I actually just found an old ticket in our backlog.  It looks like whomever generated the password in that users.conf file did not use admin:first for blake.
Did you say that you turned on file authentication using that config and you were able to log in as blake using `admin:first` as the password?
Really what we should probably do is use the correct sha in that users.conf file.  I can try to spend a bit of time verifying this next week.

MrRagnar:
Hello!
This is how my ```authentication.conf``` looks:
```
authentication {
  type = file
  file {
    file.userfile = "conf/users.conf"
  }
}
```
I have tried using both ```conf/users.conf``` and ```users.conf``` as the value for ```file.userfile```

And here is my ```users.conf```:
```
pekka:{SHA}d/F1MzIPfo+jLyxMx/TLH7ZGjkg=:engineering,infra,ops
blake:{SHA}Xziwl2ZuU/FOR4XXbjOfN7MMRNw=:engineering,infra,ops
matt:{SHA}dvxs8ZEUa332uMS+IaPl8PQ8lC4=:engineering,management
test:{SHA}nmZcGi9sch6TRiyn5uSMHgw5/Jg=:none
joeengineer:{SHA}Pkt4VUupqicwPeggw/osIEaGRAw=:engineering
```
Using this config I am able to login using ```blake``` and ```admin:first```.
But not with the account ```pekka```

Primer42:
Sorry, I'm a bit confused.  How did you generate the password sha for pekka?

MrRagnar:
I used ```htpasswd``` with the options ```-s -n```.
[...]

PR #435, opened 2016-04-06
Make blake's password 'admin:first' in the example users.conf files

Primer42:
This makes testing file authentication much easier, since it keeps
blake's password the same in all auth methods

@tumblr/collins

### DIFF
--- a/conf/docker/users.conf
+++ b/conf/docker/users.conf
@@ -1,4 +1,4 @@
-blake:{SHA}Xziwl2ZuU/FOR4XXbjOfN7MMRNw=:engineering,infra,ops
-matt:{SHA}dvxs8ZEUa332uMS+IaPl8PQ8lC4=:engineering,management
-test:{SHA}nmZcGi9sch6TRiyn5uSMHgw5/Jg=:none
-joeengineer:{SHA}Pkt4VUupqicwPeggw/osIEaGRAw=:engineering
+blake:{SHA}0sJmRmPZiSG3mjLYl0Zcd1M1W7I=:engineering,infra,ops
+matt:{SHA}iEPX+SQWIR3p67lj/0zigSWTKHg=:engineering,management
+test:{SHA}wl9ZhfKrY7rrJAii19vHnY8p0C8=:none
+joeengineer:{SHA}ymZn9lDgRKU/m3cNg1Lm5CjEtGQ=:engineering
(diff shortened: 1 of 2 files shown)


roymarantz:
:+1:

byxorna:
I dont follow... isnt the password already 'admin:first'? I know a simple `docker run -p 9000:9000 tumblr/collins` lets me log in with `blake` `admin:first`.

Primer42:
@byxorna - `default` auth is not file auth - it's a weird in-memory database setup. 

https://github.com/tumblr/collins/blob/master/conf/authentication.conf#L3
http://tumblr.github.io/collins/configuration.html#authentication
https://github.com/tumblr/collins/blob/master/app/collins/util/security/MixedAuthenticationProvider.scala#L18

Take a look at [this issue comment](https://github.com/tumblr/collins/issues/422#issuecomment-206418780) for a full explanation, but the hash in the example `users.conf` files does not actually match the hash of `admin:first`

Primer42:
Actually, I just looked at the mock authentication provider - it doesn't seem like any of the examples in users.conf match.  I'm going to make all of them match, so I don't think we know what these passwords are anymore.

Primer42:
Thinking about it, I'm not sure if this is the right approach.  We may be better off having different passwords, and documenting what they are, so that people can turn on file auth and immediately know if it's working or if its still hitting the mock auth provider.  I think both options have benefits and downsides, and I'm not sure which is best....

byxorna:
@Primer42 what i think is more likely is that the algorithm that your X year newer htpasswd is using is different than what was used when generating these files originally.

Primer42:
@byxorna see [the comment I linked to](https://github.com/tumblr/collins/issues/422#issuecomment-206418780) - it's not doing anything complicate (base64 of a sha1), I tested with the raw scala code.  And I turned on file auth with this file, and I could not log in as blake.  This is definitely a different password.

Primer42:
**bump**

byxorna:
I like having some well documented users/pws on the getting started wiki (and make clear the roles they are in as well). I.e. `root`, `opsuser`, `joeuser`, etc.

Primer42:
Hahaha
http://tumblr.github.io/collins/index.html#quickstart
The passwords are documented, in Authentication in that section.
So I think what we need to do is just make the docs better, and put this in a more logical place.

PR #440, opened 2016-04-13
add /usr array of java_home locations

davidblum:
open-jdk-1.8 for centos 6 installs into `/usr/bin/java`.  Adding `/usr` to the JAVA_HOME array lets it find it easily.


Primer42:
LGTM
@tumblr/collins - any thoughts about this?

roymarantz:
It does JAVA_HOME=/usr is correct for that RPM :+1:

PR #447, opened 2016-05-04
let your $PATH do all the work

byxorna:
Make the out of the box config work if you have ipmitool in another place in your path.

@Primer42 @schallert @defect @roymarantz

### DIFF
--- a/support/ruby/consolr/lib/consolr/runners/ipmitool.rb
+++ b/support/ruby/consolr/lib/consolr/runners/ipmitool.rb
@@ -4,11 +4,7 @@ module Consolr
   module Runners
     class Ipmitool < Runner
       def initialize config
-        @ipmitool = if config.empty?
-                      '/usr/bin/ipmitool'
-                    else
-                      config
-                    end
+        @ipmitool = config.empty? ? 'ipmitool' : config
       end
 
       def can_run? node


schallert:
LGTM. Unless we think there's other changes we want to pull in we can probably bump the version here as well, or we can do that on our end before we publish.

Primer42:
LGTM - yeah it would be nice if you bumped the version.

roymarantz:
I'm not a big fan of this.  The default is a pretty standard place for ipmitool and if you want something special, you can, and IMO, should hack the config.  I don't feel that strongly about it, so if I'm in the minority that ignore my dislike.  In general I prefer having the path specified so "everyone" gets the same value.

byxorna:
@Primer42 i wanted the upstream maintainers to make the decision about when to release a new version :)

@roymarantz i would rather not make assumptions about where a tool is located. For example, installing `ipmitool` via brew in OS X places it in `/usr/local/bin/ipmitool`. The $PATH is there exactly to provide this looseness

roymarantz:
@byxorna while that is true, my preference is for the config to be the source of truth about rather than my personal environment.  I've found that my environment often differers from other people's and even each project needs a different environment.  We are not likely to convince each other to change our minds, so we can continue to differ :-)  As I said, I don't have an issue if this is merged or not.

byxorna:
@roymarantz you can still override in the config. This change only makes the _default_ executable sans path. You are still free to set in the config whatever path to a binary you would like.

byxorna:
@Primer42 bump

PR #453, opened 2016-07-18
Consolr lazy load runners

Primer42:
That is, wait until we are going to use a runner before we load it.

When we run consolr a lot on one server, the server will run out of resources and lock up.

We think loading runners with lots of dependencies takes a lot of resources, so I want to try to load
the runners we use more often earlier, and not load bigger, less used runners if they aren't needed.

@defect @roymarantz @sushruta @schallert @qx-xp

### DIFF
--- a/support/ruby/consolr/lib/consolr.rb
+++ b/support/ruby/consolr/lib/consolr.rb
@@ -97,36 +97,9 @@ module Consolr
         abort("Please pass either the hostname OR the tag but not both.")
       end
 
-      if options[:runner].nil?
-        runners = load_runners(@config_params.fetch('runners', []))
-        # Default to the ipmitool runner for backwards compatibility
-        if runners.empty?
-          require 'consolr/runners/ipmitool'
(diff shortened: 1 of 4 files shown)


support/ruby/consolr/lib/consolr.rb
```diff
@@ -137,54 +110,83 @@ def start options
...
-        puts runner.sensors @node
-      when options[:get_sol_info]
-        puts runner.sol_info @node
-      else
-        begin
-          puts "specify an action"
-          exit 1
-        end
+      # use the command line runner, if it was provided
+      runner_names = [ options[:runner]].compact
```
schallert: [picky] extra whitespace

schallert:
LGTM. Have you tested this under various config file cases (i.e. no runner config, certain orders of runners, etc.)?

Primer42:
Yep, I've tested all of those cases.

roymarantz:
👍

PR #461, opened 2016-09-16
Fix docs for solr changes introduced in PR #268

Primer42:
@tumblr/systems


Primer42:
Or actually I should problem mention @tumblr/collins

Issue #466, opened 2016-09-19
Collins should handle missing vlan-name in LLDP reports


byxorna:
I had to hack genesis to inject vlan names when the generated LLDP report is missing them (dependent on the kind of switches attached?). It would be nice if Collins handles this automatically, and sets a default vlan name and vlan ID if missing, so it can accept lldp reports from more types of switches.

```
      log "Generating LLDP report"
      network_report = self.generate_network_report
      log "Validating all network interfaces are listed in LLDP report"
      self.validate_num_interfaces(network_report)

      lldp_xml = Nokogiri::XML(network_report)
      # aristas dont set the vlan name, and collins barfs with:
      # Exception Class: class collins.util.parsers.AttributeNotFoundException Exception Message: Found empty vlan name Message: Exception updating asset
      log "Making sure all reported vlans have names"
      set_missing_vlan_names!(lldp_xml)
```

and 

```
def self.set_missing_vlan_names!(lldp_xml)
    lldp_xml.css('lldp interface vlan').each do |node|
      if node.content.empty?
        log "VLAN name missing! Setting to default so collins accepts this LLDP report"
        node.content = "pingu"
      end
    end
    lldp_xml
end
```

PR #469, opened 2016-09-19
Add documentation for requireVlanName parameter

defect:
There is a `requireVlanName` parameter in the `lldp` namespace that allows LLDP imports where the VLAN name isn't set (this is the case for some Arista switches for instance). It's not documented anywhere else than the reference configs though, so this PR adds a quick section about it. See #466 as well.

@tumblr/collins @byxorna


roymarantz:
👍

PR #470, opened 2016-09-20
External mysql db support

schmitmd:


discordianfish:
Hi @schmitmd,

can you provide some more context for this PR?
It looks like you changed the config and init script so it works for you, but it probably breaks the setup for someone else.

schmitmd:
The intent behind this was to add support for configuring an external MySQL database server for Collins as opposed to assuming that Collins host is also the database host.  

Which part of my code would break things for someone else?

discordianfish:
@schmitmd You're right, it looks like this wouldn't change the existing behavior. LGTM.

@tumblr/collins Ping? Does this mention actually works?

schmitmd:
@tumblr/collins Ping again?  It looks like an issue with an API hit getting the wrong response.  Maybe a retry will fix it...

scripts/collins.sh
```diff
@@ -32,7 +32,18 @@ JMX_OPTS="-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.port=333
 DEBUG_OPTS="-XX:ErrorFile=${LOG_HOME}/$APP_NAME/java_error%p.log -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp/collinsDump.hprof"
 
 # Check for config overrides
-[ -f /etc/sysconfig/collins ] && . /etc/sysconfig/collins
+DISTRO=$(lsb_release -d | awk '{print $2}')
```
byxorna: this will explode if you dont have lsb-core installed. Most systems have, but cant depend on this on some RHELs. Can you fall back to processing /etc/system-release instead? /etc/os-release would be dope, but its systemd-dependent.

scripts/collins.sh
```diff
@@ -32,7 +32,18 @@ JMX_OPTS="-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.port=333
...
+
+case $DISTRO in
+    Ubuntu|Debian)
```
byxorna: this should probably be default cause honestly, cause other OSes (arch, etc) use this pattern as well

byxorna:
@schmitmd can you merge master in and push again. Lots of fixes have been applied to the unit testing infra since you opened this diff; the unit tests were pretty flaky before

scripts/collins.sh
schmitmd: I realized: Why do I need to check OS version at all?  Just source those files if they exist, right?

byxorna: 👍

byxorna:
@schmitmd its passing now :)
@defect any reservations about this landing?

schmitmd:
@defect  Any update?

schmitmd:
I got one approval, but it looks like this PR got lost in the shuffle.  Been quite a while.  Anyone willing to merge it in?

yahoocla:
Thank you for submitting this pull request, however I do not see a valid CLA on file for you. Before we can merge this request please visit https://yahoocla.herokuapp.com/ and agree to the terms. Thanks! :smile:

schmitmd:
CLA signed.

defect:
Hey @schmitmd! Very sorry for the delay. I'm cutting a ticket for myself in our internal task tracking system to take a look at this before the end of the week.

schmitmd:
End of the week was some weeks ago. You can just close this if it's not going to happen. No worries.

schmitmd:
Giving up.

PR #472, opened 2016-09-26
Collins no longer supports java 1.6

Primer42:
I tried to build with 1.6, and we are relying upon java.nio.file.Path which was
introduced in 1.7

@tumblr/collins


Primer42:
@roymarantz fixed the JCE link for java 7

PR #473, opened 2016-09-26
Ensure that we build only with java 1.7

Primer42:
Every time I release collins, I build it with the wrong version of java.

I'm updating the package.sh script to require that you build with 1.7, just so I don't keep making this mistake on releases.

@tumblr/collins

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ RUN apt-get update && \
     unzip -q ./activator.zip && \
     cd collins && \
     java -version 2>&1 && \
-    PLAY_CMD=/build/activator-$ACTIVATOR_VERSION-minimal/activator ./scripts/package.sh && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN apt-get update && \
     unzip -q ./activator.zip && \
     cd collins && \
     java -version 2>&1 && \
```
roymarantz: this java -version check seems weird/wrong since it will ignore the rest of the command instead of aborting

Primer42: Yeah, I think that was just added for debugging purposes.  @byxorna would be able to tell us for sure though 😄

Primer42:
**bump** @tumblr/collins

byxorna:
LGTM, but i apparently cant give a 👍 ?

PR #482, opened 2016-10-28
Return NODECLASS in API

discordianfish:
Hi,

the API doesn't return the nodeclass right now but I want to use that to determine different partition schemes in my installer. I think this isn't a uncommon use case. What do you think?

### DIFF
--- a/app/collins/models/asset/AllAttributes.scala
+++ b/app/collins/models/asset/AllAttributes.scala
@@ -73,6 +73,7 @@ case class AllAttributes(
   def toJsValue(): JsValue = {
     val outSeq = Seq(
       "ASSET" -> asset.toJsValue,
+      "CLASSIFICATION" -> Json.toJson(asset.nodeClass),
       "HARDWARE" -> lshw.toJsValue,
       "LLDP" -> lldp.toJsValue,
       "IPMI" -> Json.toJson(ipmi),


discordianfish:
This is btw the output for tumblrtag1 after adding a empty classifier:

```
    },
    "NODECLASS": {
      "ID": 2,
      "TAG": "app-classifier",
      "STATE": null,
      "STATUS": "Allocated",
      "TYPE": "CONFIGURATION",
      "CREATED": "2016-10-28T19:29:43",
      "UPDATED": "2016-10-28T19:30:53",
      "DELETED": null
    },
    "HARDWARE": {
```

I personally only need the tag, so could also use that instead if you prefer.
I'll also submit a PR to update the go bindings if we agree on adding this.

byxorna:
@discordianfish I think calling it something more specific than `NODECLASS` may help reduce confusion. In earlier work, I tried to move naming this feature to "asset classification" to explicitly differentiate it from the `NODECLASS` attribute that collins uses for your provisioning profile. 

I like this though! Ive wanted this feature for ages, never thought it would be this easy to add it :) (cc @alex-laties)

discordianfish:
I used nodeclass because the attribute I have to set on the classifier is IS_NODECLASS. But I'm happy to change it to whatever you prefer. What about "classification"? And should we keep all elements?

discordianfish:
@byxorna / @alex-laties  Can we get this merged?

byxorna:
@discordianfish could you change it to "CLASSIFICATION" before merging?

discordianfish:
@byxorna Done!

byxorna:
:shipit: (cc @Primer42)

michaeljs1990:
anyway to get this bumped? This would help me stop severe and everlasting damage from befalling our puppet repo.

byxorna:
cc @roymarantz @michaeljs1990 

This imho should get unit tests illustrating basic behavior validation, as well as a bit in the docs describing this thing. @discordianfish are you willing to make those changes? (i.e. at least illustrate no classification serializes to `null`).

discordianfish:
I'm not using collins anymore, so won't have time for this unfortunately.

michaeljs1990:
@discordianfish thanks for the reply i'll take this over then just didn't want to steal it from you.

PR #487, opened 2016-11-08
Document changes around encrypted tags permissions.

Primer42:
- Addition of canWriteEncryptedTags
- Renaming of canSeePasswords to canSeeEncryptedTags

Code PR: https://github.com/tumblr/collins/pull/486

### DIFF
--- a/_data/api.yml
+++ b/_data/api.yml
@@ -97,6 +97,7 @@ sections:
         description: Update an asset
         request: POST /api/asset/:tag
         permission: controllers.AssetApi.updateAsset
+        note: If you are setting an encrypted tag, you must also have the features.canWriteEncryptedTags permission. See the <a href="configuration.html#features">features documentation</a> for more information.
         parameters:
           lshw:
             type: Option[String]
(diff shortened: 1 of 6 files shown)


byxorna:
Where are the docs for `canWriteEncryptedTags`? I think breaking the changes to _includes/configuration/features.html into 2 separate bullets would make things clearer?

Primer42:
I thought about this too while I was writing up these changes.
Technically, `_includes/configuration/features.html` is for documenting the features, i.e. the `encrypted_tags` configuration, not permissions. We also say [here](https://github.com/tumblr/collins/blob/7bd4fa07c13447b51d590c2f108f2d1d50d25308/_includes/configuration/authentication.html#L30) that the permissions are documented under the api endpoints that use each permission.
As such, I added docs to the api page about the new permissions I added:
https://github.com/tumblr/collins/pull/487/files#diff-575a803dd8f76bb5f51928e633cd8548R100

I agree, it isn't very clear.  But I think that's mostly because the docs around permissions are not very clear.
I would be OK adding a new section that documents each permissions and what it does separately from the API docs, or doing something else to try to document permissions better.

PR #488, opened 2016-11-10
The permissions are named `feature.` not `features.`

Primer42:
@tumblr/collins RFR


roymarantz:
if this is correct, then 👍

PR #492, opened 2016-11-28
Skip UID verification when ipmi config is missing

discordianfish:
This will make collins' intake process skip UID based identification when IPMI data is missing.

Let me know if you need this configurable.

### DIFF
--- a/app/collins/util/config/Feature.scala
+++ b/app/collins/util/config/Feature.scala
@@ -30,6 +30,7 @@ object Feature extends Configurable {
   def encryptedTags = getStringSet("encryptedTags")
   def keepSomeMetaOnRepurpose = getStringSet("keepSomeMetaOnRepurpose", Set())
   def intakeSupported = getBoolean("intakeSupported", true)
+  def intakeIpmiOptional = getBoolean("intakeIpmiOptional", false)
   def ignoreDangerousCommands = getStringSet("ignoreDangerousCommands")
   def hideMeta = getStringSet("hideMeta")
   def noLogAssets = getStringSet("noLogAssets")
--- a/app/collins/controllers/actions/resources/FindAction.scala
+++ b/app/collins/controllers/actions/resources/FindAction.scala
@@ -7,9 +7,12 @@ import collins.controllers.actions.ActionHelper
 import collins.controllers.actions.RequestDataHolder
 import collins.controllers.actions.asset.AssetFinderDataHolder
 import collins.controllers.actions.asset.{ FindAction => AssetFindAction }
+import collins.models.IpmiInfo
+import collins.models.Asset
 import collins.models.asset.AssetView
 import collins.models.shared.Page
 import collins.models.shared.PageParams
+import collins.util.config.Feature
 import collins.util.security.SecuritySpecification
 
 case class FindAction(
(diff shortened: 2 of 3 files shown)


roymarantz:
Shouldn't there be a config "feature" test enabling this change in behavior?  Ideally there would be unit tests too.

discordianfish:
@roymarantz 'config "feature" test'? You mean adding some flag to `app/collins/intake/IntakeConfig.scala` like strict_validation:

- default: true, current behaviour
- false, skip validation steps for which config is missing (like UID for ipmi but later maybe chassis tag if that's missing)

Tests would be ideally but I'm not very familiar with scala and there are no existing tests for the intake stuff I could extend easily. If you can point me in the right direction, I can look into this though.

roymarantz:
By feature I mean something like https://tumblr.github.io/collins/configuration.html#features and a simple example of its use can be found in app/collins/util/StateMachines.scala just look for Feature.
I can't find an instance of a similar kind of test 😞 
Is there an advantage to adding such a feature test in app/collins/intake/IntakeConfig.scala instead of where your code change is?

discordianfish:
Ah, i see. Yes such feature works for me. Would call it sloppyIntake?

roymarantz:
The name doesn't matter to much, but I'd favor something more descriptive like optionalIpmiIntake

discordianfish:
@roymarantz Like this?

I also want to add another feature to allow skipping chassis verification as well, so might want to wait before merging this.

roymarantz:
this looks fine with me 👍 
I'd suggest adding the other feature in another PR unless you believe it is tightly linked to this one.

discordianfish:
Fine with me. Then this is ready to get merged!

app/collins/util/config/Feature.scala
```diff
@@ -30,6 +30,7 @@ object Feature extends Configurable {
   def encryptedTags = getStringSet("encryptedTags")
   def keepSomeMetaOnRepurpose = getStringSet("keepSomeMetaOnRepurpose", Set())
   def intakeSupported = getBoolean("intakeSupported", true)
+  def intakeIpmiOptional = getBoolean("intakeIpmiOptional", false)
```
discordianfish: How about this? I'm still pushing the gh-pages changes. Unfortunately that branch is huge.. :-/ Another TODO I guess..

For now, what about these doc changes?

```
commit 6c03d6e7e6ac0b0ea6a2c0f18168a5ee796ec5af
Author: Johannes 'fish' Ziemke <github@freigeist.org>
Date:   Fri Dec 2 19:06:50 2016 +0100

    Document intakeIpmiOptional

diff --git a/_includes/configuration/features.html b/_includes/configuration/features.html
index 446a041..5cf2bdd 100644
--- a/_includes/configuration/features.html
+++ b/_includes/configuration/features.html
@@ -72,6 +72,11 @@ feature flags are documented below.
       <td>Whether the web based intake is supported or not</td>
     </tr>
     <tr>
+      <td class="inlinecode">intakeIpmiOptional</td>
+      <td>Boolean</td>
+      <td>Whether the tumblr intake process may skip UID verification for assets without IPMI configuration.</td>
+    </tr>
+    <tr>
       <td class="inlinecode">keepSomeMetaOnRepurpose</td>
       <td>Set[String]</td>
       <td>If useWhitelistOnRepurpose is true, this is a list of attributes to keep</td>
```

discordianfish:
@Primer42 Ping?

PR #493, opened 2016-12-02
Document intakeIpmiOptional

discordianfish:
Belongs to #492

### DIFF
--- a/_includes/configuration/features.html
+++ b/_includes/configuration/features.html
@@ -72,6 +72,11 @@ feature flags are documented below.
       <td>Whether the web based intake is supported or not</td>
     </tr>
     <tr>
+      <td class="inlinecode">intakeIpmiOptional</td>
+      <td>Boolean</td>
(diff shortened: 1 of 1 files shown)


_includes/configuration/features.html
```diff
@@ -72,6 +72,11 @@
       <td>Whether the web based intake is supported or not</td>
     </tr>
     <tr>
+      <td class="inlinecode">intakeIpmiOptional</td>
+      <td>Boolean</td>
+      <td>Whether the tumblr intake process may skip UID verification for assets without IPMI configuration.</td>
```
roymarantz: I don't think tumblr should be here.

roymarantz: I only found a reference in collins/conf/reference/features_reference.conf, but not a big deal either way, even though it should be reworked too.

discordianfish:
Ok, if the plan is to change it everywhere else it makes sense to change it here. Did that and pushed it.

PR #494, opened 2016-12-06
make default docker permissions.yaml synchronized with latest changes

byxorna:
@Primer42 @defect @roymarantz @qx-xp


roymarantz:
👍

PR #496, opened 2016-12-09
Allow skipping chassis tag verification

discordianfish:
With the feature intakeChassisTagOptional enabled, collins will skip the
chassis tag verification in the intake process if the chassis tag isn't
set. It will allow adding the asset tag in the final intake step.

This is based on top of #492

### DIFF
--- a/app/collins/controllers/actions/resources/FindAction.scala
+++ b/app/collins/controllers/actions/resources/FindAction.scala
@@ -7,9 +7,12 @@ import collins.controllers.actions.ActionHelper
 import collins.controllers.actions.RequestDataHolder
 import collins.controllers.actions.asset.AssetFinderDataHolder
 import collins.controllers.actions.asset.{ FindAction => AssetFindAction }
+import collins.models.IpmiInfo
+import collins.models.Asset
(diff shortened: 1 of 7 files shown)


byxorna:
Eyeballing it lgtm, but i havent tested all the different configuration modes.

discordianfish:
@byxorna Best would be to add some tests.. but since there are none for the intake process so far and I'm not very experienced with scala I'd need some help with that.

byxorna:
@discordianfish the testing framework is kind of complicated, so I dont have much guidance here. @Primer42 ?

Primer42:
Yeah, similar review to @byxorna - code looks good, but I haven't tested it thoroughly. 
There aren't any tests around intake, and I think it's out of scope to ask you to write a big part of the test framework as part of this PR.

discordianfish:
Ok, then let's test this manually. I've tried it locally and deployed it to our systems. So far works as expected but our setup is tiny and the intake process not yet much used.

discordianfish:
Looks like this project is dead outside tumblr at least, so closing this.

PR #498, opened 2016-12-15
Only query allocated remote assets

Primer42:
If you decommission an asset for multicollins, it shouldn't try to query it anymore.

I have tested this and it works as expected.

@tumblr/collins @byxorna @discordianfish

### DIFF
--- a/app/collins/models/Asset.scala
+++ b/app/collins/models/Asset.scala
@@ -255,7 +255,7 @@ object Asset extends Schema with AnormAdapter[Asset] with AssetKeys {
   def findMulti(page: PageParams, params: AttributeResolver.ResultTuple, afinder: AssetFinder, operation: Option[String], details: Boolean): Page[AssetView] = {
     val instanceFinder = AssetFinder(
       tag = None,
-      status = None,
+      status = Status.Allocated,
       createdAfter = None,
       createdBefore = None,
       updatedAfter = None,


discordianfish:
@Primer42 I never used multicollins (but we're planning to), so not sure about the details. But why limit it to allocated assets? Shouldn't it possible to search for any kind of asset, even decomissioned?

roymarantz:
I believe this code is used to find the remote collins instances to query so it 👍

Primer42:
@discordianfish sorry for the confusion - the way multicollins works is that you create an asset (usually of type DATA_CENTER) in your local instance of collins, and set an attribute (default is LOCATION) with the url with creds that allow you to query the remote collins i.e. http://user:pass@collins.other-dc.example.com/.  If you ask collins to do a remote query, it searches its local database for all of the assets of the correct type with the LOCATION attribute set, and uses the value of that LOCATION attribute to query the remote collins, and returns the results all together.
I'm modifying the code that does that local search, so when it's trying to find the remote datacenter assets, it only chooses Allocated assets instead of choosing all datacenter assets regardless of state.
Does that make more sense?

ssgelm:
Just FYI this should be documented somewhere. Nowhere is it specified in the docs that DATA_CENTER assets must be in the allocated state and ours were not (cause it never seemed to matter in the past).

PR #501, opened 2016-12-19
Fix "respond_to? is old fashion" warning

Primer42:
Ruby 2.3 has started complaining if your respond_to? doesn't take a 2nd argument

@tumblr/collins @byxorna


byxorna:
LGTM

roymarantz:
sigh. 👍

PR #504, opened 2017-01-15
Fix relative path to collins-client in collins-shell bin

ssgelm:
This was broken in commit 27a81210cf2a8affb4a2c489a53b696bec18f13c.


roymarantz:
@ssgelm thanks for the patch & sorry for the delay.  We'll cut a new release in a little while.

Issue #389, opened 2015-12-17
Ability to delete a specific IP address


ssgelm:
For better or for worse we have a situation where assets somewhat regularly have multiple IP addresses in the same pool.  Ideally we would like a way to delete a single IP from an asset via the API.  Is that reasonable?  If you think it is I am happy to write the code, just making sure that it is functionality you this would be useful.

byxorna:
@ssgelm that would be phenomenal! :+1:
I know I have a handful of use cases that would really benefit from individually management of IPs on an asset, but never got around to implementing. If you would do the leg work, we would gladly accept it!

ssgelm:
Awesome, I'll take a stab at it!

ssgelm:
Well, a year later than I intended but I implemented this in #505.

ssgelm:
This was done in #505.

PR #505, opened 2017-01-15
Delete single ip address

ssgelm:
As discussed in #389, this adds support for deleting a single IP address via the api.  Support is also added to `collins-shell` and `collins-client` for the newly added api call.  I was looking to add the new `address` parameter to the API docs too but it looks like that only lives in the `gh-pages` branch so that will require a separate PR.

### DIFF
--- a/app/collins/controllers/actions/ipaddress/DeleteAction.scala
+++ b/app/collins/controllers/actions/ipaddress/DeleteAction.scala
@@ -3,6 +3,7 @@ package collins.controllers.actions.ipaddress
 import scala.concurrent.Future
 
 import play.api.data.Form
+import play.api.data.Forms.tuple
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 import play.api.libs.json.JsNumber
 import play.api.libs.json.JsObject
@@ -22,26 +23,32 @@ case class DeleteAction(
   assetTag: String,
   spec: SecuritySpecification,
   handler: SecureController
-) extends SecureAction(spec, handler) with AssetAction with ParamValidation {
+) extends SecureAction(spec, handler) with AssetAction with ParamValidation with AddressActionHelper {
 
-  case class ActionDataHolder(asset: Asset, pool: Option[String]) extends RequestDataHolder
+  case class ActionDataHolder(asset: Asset, pool: Option[String], address: Option[String]) extends RequestDataHolder
 
-  val dataForm = Form(
-    "pool" -> validatedOptionalText(1)
-  )
+  val dataForm = Form(tuple(
+    "pool" -> validatedOptionalText(1),
+    "address" -> validatedOptionalText(1)
+  ))
 
   override def validate(): Validation = withValidAsset(assetTag) { asset =>
-    val pool: Option[String] = dataForm.bindFromRequest()(request).fold(
-      err => None,
-      str => str
-    )
-    Right(ActionDataHolder(asset, pool))
+    val (pool, address): Tuple2[Option[String], Option[String]] = dataForm.bindFromRequest()(request).get
+    address match {
+      case Some(ip: String) => withValidAddress(ip) { validAddress => 
+        Right(ActionDataHolder(asset, pool, address))
+      }
+      case None => Right(ActionDataHolder(asset, pool, None))
+    }
   }
 
   override def execute(rd: RequestDataHolder) = Future {
     rd match {
-      case ActionDataHolder(asset, pool) =>
-        val deleted = IpAddresses.deleteByAssetAndPool(asset, pool)
+      case ActionDataHolder(asset, pool, address) =>
+        val deleted = address match {
+          case Some(ip: String) => IpAddresses.deleteByAssetAndAddress(asset, ip)
+          case None => IpAddresses.deleteByAssetAndPool(asset, pool)
+        }
         tattler.notice("Deleted %d IP addresses".format(deleted), asset)
         ResponseData(Status.Ok, JsObject(Seq("DELETED" -> JsNumber(deleted))))
     }
--- a/app/collins/controllers/actions/ipaddress/DeleteAction.scala
+++ b/app/collins/controllers/actions/ipaddress/DeleteAction.scala
@@ -49,7 +49,10 @@ case class DeleteAction(
           case Some(ip: String) => IpAddresses.deleteByAssetAndAddress(asset, ip)
           case None => IpAddresses.deleteByAssetAndPool(asset, pool)
         }
-        tattler.notice("Deleted %d IP addresses".format(deleted), asset)
+        address match {
+          case Some(ip: String) if deleted > 0 => tattler.notice("Deleted IP address %s".format(address), asset)
+          case _ => tattler.notice("Deleted %d IP addresses".format(deleted), asset)
+        }
         ResponseData(Status.Ok, JsObject(Seq("DELETED" -> JsNumber(deleted))))
     }
   }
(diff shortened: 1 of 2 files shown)


ssgelm:
In lieu of the proper documentation (for which I can make a PR to `gh-pages` if this looks good to you all) this adds an `address` parameter to the `DELETE  /api/asset/:tag/addresses` route.  If `address` is passed, the address is deleted assuming it is currently allocated to the provided tag.

app/collins/controllers/actions/ipaddress/DeleteAction.scala
```diff
@@ -22,26 +23,32 @@ case class DeleteAction(
...
   override def execute(rd: RequestDataHolder) = Future {
     rd match {
-      case ActionDataHolder(asset, pool) =>
-        val deleted = IpAddresses.deleteByAssetAndPool(asset, pool)
+      case ActionDataHolder(asset, pool, address) =>
+        val deleted = address match {
+          case Some(ip: String) => IpAddresses.deleteByAssetAndAddress(asset, ip)
+          case None => IpAddresses.deleteByAssetAndPool(asset, pool)
+        }
         tattler.notice("Deleted %d IP addresses".format(deleted), asset)
```
roymarantz: this should reflect that the single address (vs all addresses of the asset) have been deleted

support/ruby/collins-client/lib/collins/api/ip_address.rb
```diff
@@ -33,11 +33,12 @@ def ipaddress_update! asset_or_tag, old_address = nil, options = {}
       end
     end
 
-    def ipaddress_delete! asset_or_tag, pool = nil
+    def ipaddress_delete! asset_or_tag, pool = nil, address = nil
       asset = get_asset_or_tag asset_or_tag
       logger.debug("Deleting addresses for asset #{asset.tag} in pool #{pool}")
```
roymarantz: please log differently if all addresses for the asset or a particular one

roymarantz:
This looks real good.  I'd like to get the documentation & logging setup before merging though.  thanks

support/ruby/collins-client/lib/collins/api/ip_address.rb
```diff
@@ -33,11 +33,12 @@ def ipaddress_update! asset_or_tag, old_address = nil, options = {}
       end
     end
 
-    def ipaddress_delete! asset_or_tag, pool = nil
+    def ipaddress_delete! asset_or_tag, pool = nil, address = nil
```
byxorna: This feels odd, in that if you want to pass an address, you have to pass a nil pool? Feels like a better use of an options hash to me, where pool and address are pushed into the final hash param

byxorna:
This is awesome! Once @roymarantz's comments have been addressed, lgtm!

support/ruby/collins-client/lib/collins/api/ip_address.rb
ssgelm: Good idea - I'll refactor that.  My initial intent was to minimize other code changes but I think an options hash makes more sense.

support/ruby/collins-client/lib/collins/api/ip_address.rb
ssgelm: Sounds good will do!

app/collins/controllers/actions/ipaddress/DeleteAction.scala
ssgelm: Noted.  Will fix!

ssgelm:
Docs change PRed in #508

ssgelm:
@roymarantz and @byxorna I have pushed new commits to fix your concerns.

support/ruby/collins-client/lib/collins/api/ip_address.rb
```diff
@@ -33,12 +33,12 @@ def ipaddress_update! asset_or_tag, old_address = nil, options = {}
       end
     end
 
-    def ipaddress_delete! asset_or_tag, pool = nil, address = nil
+    def ipaddress_delete! asset_or_tag, options = {}
       asset = get_asset_or_tag asset_or_tag
-      logger.debug("Deleting addresses for asset #{asset.tag} in pool #{pool}")
+      logger.debug("Deleting addresses for asset #{asset.tag} in pool #{options[:pool]} with address #{options[:address]}")
```
roymarantz: will a missing key (say :address) print correctly or do you need to handle that better?

roymarantz:
as I said in [#508](https://github.com/tumblr/collins/pull/508) should the code exclude specifying both a pool and address?

support/ruby/collins-client/lib/collins/api/ip_address.rb
ssgelm: Ruby "helpfully" defaults missing hash keys to `nil` so this just works:
```
$ bin/collins-shell ip_address delete DEV --tag=tumblrtag1 --debug
DEBUG [2017-02-06 15:41:49.560] collins-shell: Deleting addresses for asset tumblrtag1 in pool DEV with address
```

ssgelm:
I'm not sure there's much of a risk to allow the end user to specify both.  Address takes precedence over pool which makes sense to me as it's more specific.  If a user were to specify both I can't imagine they would be surprised to see the single address deleted.

roymarantz:
OK this is 👍
Thanks

ssgelm:
Now that #508 is merged are you planning on merging this?

support/ruby/collins-client/lib/collins/api/ip_address.rb
```diff
@@ -33,11 +33,12 @@ def ipaddress_update! asset_or_tag, old_address = nil, options = {}
       end
     end
 
-    def ipaddress_delete! asset_or_tag, pool = nil
+    def ipaddress_delete! asset_or_tag, options = {}
```
ssgelm: Sure.  I believe the former is the best option there although the latter should be possible thanks to the magic of dynamic types.  Which would you prefer?

app/collins/controllers/actions/ipaddress/DeleteAction.scala
```diff
@@ -22,27 +23,36 @@ case class DeleteAction(
...
-      str => str
-    )
-    Right(ActionDataHolder(asset, pool))
+    val (pool, address): Tuple2[Option[String], Option[String]] = dataForm.bindFromRequest()(request).get
+    address match {
+      case Some(ip: String) => withValidAddress(ip) { validAddress => 
+        Right(ActionDataHolder(asset, pool, address))
+      }
+      case None => Right(ActionDataHolder(asset, pool, None))
+    }
```
ssgelm: Sure.  I didn't add any tests cause there weren't any for any of these classes already.  I'll take a stab at adding some tests.

ssgelm:
Sorry to bump this after so long but what concretely would you all like me to fix here?  Just the ruby client?  Adding unit tests seems like a nice to have but the rest of this code isn't tested yet so it seems like a good amount of work...

defect:
Hey @ssgelm! Sorry for dropping the ball on this. I'm assigning myself to this PR to get this rolling again.

defect:
We split out the ruby gems to their own repos ([collins_client](https://github.com/tumblr/collins_client) and [collins_shell](https://github.com/tumblr/collins_shell)]) not too long ago, so i think a first step would be to move those changes over to those repos, or at least remove those changes from this PR since they're blocking merging.

Personally I don't think this PR should be blocked on adding it to the client libraries as long as we don't break current functionality, though it would be real nice to have support for it there :) But I think that can happen async from this.

defect:
Again.. sorry for the delay :( This looks good to me. Let me see if i can get @byxorna 's eyes on it again since he had a few comments.

ssgelm:
Looks like everyone is ok with this... can it be merged?

defect:
Very sorry @ssgelm. 2+ years for a PR is quite embarrassing. I'll merge this tonight or tomorrow morning.

PR #506, opened 2017-01-17
Add ipmitool to container

michaeljs1990:
ipmitool tool is needed for the web ui intake process which is enabled by default. I believe it makes sense to have it installed here and also I am trying to setup everything without breaking away from the custom docker image you provide.


roymarantz:
@michaeljs1990 thanks for the patch & sorry about the delay

michaeljs1990:
🎆 🎆 🎆  no problem, thanks for landing :)

PR #507, opened 2017-01-19
Add EXTRA_OPTS for extra java options which do not fit other types

vhp:
Hi,

Due to an internal Tumblr project we are in need of adding extra flexibility to the init script. The sysconfig configuration file is handled by the users, starting with a very basic default. The need here is to allow certain java command line options which may not fit the categories otherwise specified. Most are very specific. APP_OPTS the most generic comes after sourcing of the sysconfig file so there's no much wiggle room.

@defect @roymarantz @qx-xp @komapa @byxorna @gtorre 

Thanks a bunch,

Vincent


defect:
LGTM!

PR #508, opened 2017-01-29
Update api docs to add option for deleting a single IP

ssgelm:
As added in #505.  This updates the docs to go along with that PR.

### DIFF
--- a/_data/api.yml
+++ b/_data/api.yml
@@ -1748,7 +1748,7 @@ sections:
               }
             }
       "delete an address":
-        description: Delete all addresses associated with a given tag, optionally associated with a pool
+        description: Delete addresses associated with a given tag: either all addresses, those associated with a specified pool, or a single address.
         request: DELETE /api/asset/:tag/addresses
         permission: controllers.IpAddressApi.purgeAddresses
         parameters:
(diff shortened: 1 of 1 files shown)


defect:
Awesome, thanks! 👍

_data/api.yml
```diff
@@ -1758,6 +1758,9 @@ sections:
           pool:
             type: Option[String]
             description: If specified, only addresses associated with the given pool are deleted. Otherwise, all addresses are deleted.
+          address:
+            type: Option[String]
+            description: If specified, delete only the given address.
```
roymarantz: perhaps something like:
If specified, delete only the given address, else all of them.

and modify:
description: Delete all addresses associated with a given tag, optionally associated with a pool
to be something like:
description: Delete addresses associated with a given tag, optionally associated with a pool, or optional single address

ssgelm:
@roymarantz I updated the docs as requested.  Does this look better?

_data/api.yml
```diff
@@ -1748,7 +1748,7 @@ sections:
               }
             }
       "delete an address":
-        description: Delete all addresses associated with a given tag, optionally associated with a pool
+        description: Delete addresses associated with a given tag: either all addresses, those associated with a specified pool, or a single address.
```
roymarantz: the code seem to support specifying a pool and address.  is it wrong or is this line?

ssgelm: The code should support it but I'm not sure what the value would be.

roymarantz: I think I was just misreading this, so don't worry about it.

roymarantz:
👍

PR #511, opened 2017-02-11
Add mention of internal wiki in internal docs section

vhp:
Hey,

So now on the internal wiki we have some docs on deployments specific to our environment. They are in the Collins Runbook. If there's concern about this separation of docs feel free to let me know. 

RFR @defect @roymarantz @qx-xp @komapa @gtorre

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Email collins-sm@googlegroups.com or see the mailing list archive at https://gro
 
 To create a production zip and deploy to production:
 
+  - Check Collins Runbook on internal wiki for environment specific considerations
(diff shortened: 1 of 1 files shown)


roymarantz:
IIRC you shouldn't mention internal stuff on github.com, but you can mention github.com stuff on internal stuff.  You could check with the OpenSource group to double check.

vhp:
I understand. I guess this was no more than telling folks to check 
internal wiki. Seems pretty vanilla. We don't want folks using the wrong 
directions.


roymarantz wrote on 02/27/2017 09:57 AM:

Issue #403, opened 2016-01-27
Embedded solr never reindexes on startup


byxorna:
When using embedded solr, it appears that Collins will never trigger a full reindex on startup if the collins index is missing. You can see that it initializes the index, but never performs the repopulation (i do that manually by hitting `/admin/solr`). This is an issue when running collins in a container with a separate mysql instance. Each time the container is recreated, a reindex should be triggered (otherwise searching will return [] which is super undesirable).

```
2016-01-27 18:53:40,851 - [INFO] - c.u.c.Registry$ - p.a.LoggerLike$class:info:93 - Initializing configuration of typeclass collins.util.security.LdapAuthenticationProviderConfig$
2016-01-27 18:53:40,852 - [INFO] - c.u.c.Registry$ - p.a.LoggerLike$class:info:93 - Initializing configuration of typeclass collins.util.views.TagDecoratorConfig$
2016-01-27 18:53:40,855 - [INFO] - c.u.c.Registry$ - p.a.LoggerLike$class:info:93 - Initializing configuration of typeclass collins.intake.IntakeConfig$
2016-01-27 18:53:40,856 - [INFO] - c.u.c.Registry$ - p.a.LoggerLike$class:info:93 - Initializing configuration of typeclass collins.graphs.GangliaGraphConfig$
2016-01-27 18:53:40,856 - [INFO] - c.u.c.Registry$ - p.a.LoggerLike$class:info:93 - Initializing configuration of typeclass collins.graphs.FibrGraphConfig$
2016-01-27 18:53:40,858 - [INFO] - c.u.c.Registry$ - p.a.LoggerLike$class:info:93 - Initializing configuration of typeclass collins.frames.ViewsConfig$
2016-01-27 18:53:40,875 - [INFO] - c.h.HazelcastHelper$ - p.a.LoggerLike$class:info:93 - Initializing hazelcast, enabled - false
2016-01-27 18:53:40,886 - [DEBUG] - application - c.u.s.LdapAuthenticationProvider:<init>:63 - LDAP URL: ldaps://xxx.company.com
2016-01-27 18:53:41,058 - [INFO] - play - p.a.LoggerLike$class:info:93 - Starting application default Akka system.
2016-01-27 18:53:41,450 - [WARN] - o.a.s.c.SolrResourceLoader - o.a.s.c.SolrResourceLoader:addToClassLoader:191 - Can't find (or read) directory to add to classloader: lib (resolved as: /opt/collins/conf/
solr/lib).
[...]

PR #513, opened 2017-03-06
IPMI network allocations API with pool support

### LINKED ISSUE #512: Create/fix /asset/:tag/ipmi endpoint to allow generating IPMI details on existing asset
Currently IPMI details (user/pw/IP) are generated on asset creation time. There is no endpoint to "get the next unused IPMI IP" for an asset without IPMI, which is useful for high touch induction scenarios. This also can open the door for more intelligent induction processes that can select the IPMI network pool for address allocation after creating an asset.

I am working on a diff to address this now, assigning to myself.

byxorna:
We have a bit of an odd IPMI network topology, and I have been wanting to add an endpoint to allow IPMI details to be generated separate from asset creation. This diff enables multiple OOB networks to be managed, and gives users the flexibility to generate IPMI details based on information collected at induction time. For example, one could imagine an induction process in Genesis that:

- creates the asset without IPMI
- performs LLDP collection
- call `#ipmi_allocate(tag, pool: ...)` based on the connectivity discovered and specify a specific IPMI address pool

Ive updated collins-shell and collins-client with this functionality as well.

cc @michaeljs1990 @defect @roymarantz @alex-laties @vhp @qx-xp 

Fixes #512, docs in #514


michaeljs1990:
Just to keep everything in the PR. I had an oob (lol i couldn't help myself) conversation with gabe that since you can generate IPMI at asset creation time it would be cool if you could also pass it a pool to allocate from since you could have gotten LLDP info before asset creation.

byxorna:
@michaeljs1990 ive updated diff to support `#create!('tag', ipmi_pool: "FOO")`, as well as the documentation update that goes along with it. Tests are broken for a few reasons:

1. i bumped the gem dependencies for collins-shell (as it requires the latest collins-client), but that version doesnt exist in rubygems yet
2. flaky scala solr tests bailed :( idk how to tell travis to rerun though...

@defect review appreciated!

michaeljs1990:
Was messing around with the docker build and noticed the 500 error you were talking about. I dug into it a bit and tracked it down to `app/collins/util/Tattler.scala:83` which I believe is trying to grab a single asset to use for logging. Since it's currently not set I think it's returning whatever is set as the multicollins asset in the config and trying to log to that.. which is also not set. 

```
# production.conf
features {
  syslogAsset = tumblrtag1
}
```

Needs to be set so we can use the super special logger. I should make a task for documenting this and then maybe just having it log to a central place eventually.

byxorna:
@michaeljs1990 so, i think this is a race in startup with solr initial indexing; ive seen this before, and the tumblrtag1 asset (inserted here https://github.com/tumblr/collins/blob/master/conf/evolutions/collins/2.sql#L61 by evolutions) is not always indexed by solr. Manually hitting `/admin/solr` to reindex assets sorts the problem out. We should make a new issue to debug if the issue I already had open in #403 is different.

Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM java:8-jre
+FROM openjdk:8-jre
 MAINTAINER Gabe Conradi <gabe@tumblr.com>
```
roymarantz: probably should change the MAINTAINER to sre-sys@tumblr.com

Dockerfile
```diff
@@ -1,12 +1,12 @@
...
 
 # Solr cores should be stored in a volume, so we arent writing stuff to our rootfs
 VOLUME /opt/collins/conf/solr/cores/collins/data
 
 COPY . /build/collins
 RUN apt-get update && \
-    apt-get install --no-install-recommends -y openjdk-8-jdk zip unzip ipmitool && \
+    apt-get install --no-install-recommends -y openjdk-8-jdk openjdk-8-jdk-headless zip unzip ipmitool && \
```
roymarantz: isn't openjdk-8-jdk-headless already installed in the base image?

Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM java:8-jre
+FROM openjdk:8-jre
```
roymarantz: Have you though of using openjdk:8-jre-alpine ?

Dockerfile
byxorna: no, this is the JRE image. We install only the JDK for the duration of the build, then remove it

Dockerfile
byxorna: It uses muscl instead of glibc, and didnt want to take a chance of breaking something. Plus, for downstream consumers, they expect to be able to `apt-get install` stuff for collins to hook into. We could maintain 2 different Dockerfiles that produces a debian and alpine build, but probably outside the scope of this diff

Dockerfile
byxorna: Is that alias accessible from outside the company? I thought there was a public maintainers mailing list for collins?

Dockerfile
roymarantz: does line 47 of https://github.com/docker-library/openjdk/blob/445f8b8d18d7c61e2ae7fda76d8883b5d51ae0a5/8-jre/Dockerfile#L47 already install it?

byxorna: Nah, thats only JRE. We need JDK for the build

byxorna:
@roymarantz @defect thumbs bump?

byxorna:
@roymarantz @defect second bump! I would like to land this soon, along with the already approved #514 :)

conf/test.conf
```diff
@@ -1,7 +1,7 @@
-# THIS IS NOT A PRODUCTION CONFIGURATION - DEVELOPMENT ONLY
+# THIS IS NOT A PRODUCTION CONFIGURATION - UNIT TEST ONLY
 # ~~~~~
 
-include "dev_base.conf"
+include "test_base.conf"
```
michaeljs1990: aha! :)

byxorna: oops, sorry! I figured i had enough of changing things for dev and accidentally breaking tests -_-

PR #516, opened 2017-03-07
pin rake for collins-state gem

### LINKED ISSUE #510: travis-ci ruby tests broken RUN_RUBY
the travis-ci RUB_RUBY tests are failing most probably because gemspec files want spec 2.99 but that is not compatible with rake newer than 10.x and since rake's version isn't specified 12.x is installed.

Fix, either upgrade to spec 3.x or limit rake to 10.x

byxorna:
Attempting to fix #510

@roymarantz @qx-xp @defect 

Closes #510


roymarantz:
👍

PR #521, opened 2017-03-09
Add IPMI pools to /api/address/pools endpoint

### LINKED ISSUE #515: Endpoint for retrieving list of IPMI pools
With #512 hopefully being finished soon it would be helpful to have an API endpoint for retrieving 
the list of current IPMI pools available so you can programmatically inspect the pools and select one.

This could possibly be a part of `/api/address/pools` or added as a new endpoint to `/api/ipmi/pools`.

Self assigning for now although i don't think github will actually let me.

michaeljs1990:
This adds support for programmatically discovering the currently configured pools. If no pools are configured and you have one top level IPMI pool nothing is returned such as with this config.

```
 ipmi {
   randomUsername = false
   username = "root"
   passwordLength = 16
   network="172.16.32.0/20"
   startAddress="172.16.32.20"
}
```

I believe this is the right action since you likely don't care about ipmi pools if you have not configured any and makes handling edge cases easier. Such as if you had pools configured as well as having a network set in the top level of the ipmi config. Also any default name we would give the top level pool could potentially be overwritten by a user leading to some strange cases.

Closes #515 

todo
 - [x] implement api in collins shell
- [x] update gh pages in another PR. #533
 
@byxorna


app/collins/controllers/actions/ipaddress/GetPoolsAction.scala
```diff
@@ -38,14 +39,25 @@ case class GetPoolsAction(
           } else {
             IpAddresses.getPoolsInUse()
           }
-        format(pools)
+        val ipmiPools: Set[String] =
+          if (all) {
+            IpmiInfo.AddressConfig.map(_.poolNames).getOrElse(Set())
+          } else {
+            // TODO: Change this to only return active pools?
+            IpmiInfo.AddressConfig.map(_.poolNames).getOrElse(Set())
```
michaeljs1990: This is the one thing I don't like about this currently. How IpAddress pools are handing this is just querying the DB and returning all unique fields. Since this is not in the DB at all for IPMI pools we would have to run a very expensive query to figure out what all the active pools are. The other option is writing a migration to go with an evolution which I believe is the correct path forward.

byxorna: I prefer reading the configured pools from the config for IPMI tbh. However, we would need to take into account the fact that IPMI addresses can be "loose", and not adherent to any pool in the config.

I am open to either, however an evolution will definitely complicate things. I also think if we track pool with an ipmi_info entry, we will need to update it when people hit /api/asset/:tag/ipmi to create a new address or update existing...

byxorna:
When #513 lands, this will need to be rebased, as I split out the dev_base config from the test config.

I was thinking having a separate API endpoint for IPMI pools would make more sense, because the semantics of how IPMI pools and IP pools are reported/configured/tracked/updated are different. If we are only reporting the IPMI pools as configured in the play config, I would prefer them to be handled by another endpoint with a singular purpose :)

michaeljs1990:
That sounds good to me, i'll split this out into another endpoint. Do you have any preference on the name?

byxorna:
Not really sure. Perhaps similar to https://tumblr.github.io/collins/api.html#api-ipam-get-address-pools, `GET /api/ipmi/pools`? Perhaps there can be some params to that endpoint to include/omit default pool, or include/omit actual IPs grouped by pools, and hopefully some way to indicate non-managed IP pools? (maybe the last thing is a huge timesink; maybe just stub out a param to show unmanaged IPMI network utilization and we can implement it later?)

app/collins/controllers/IpmiApi.scala
```diff
@@ -85,4 +88,8 @@ trait IpmiApi {
     )
   }(Permissions.IpmiApi.UpdateIpmi)
 
+  // GET /api/ipmi/pools
+  def getIpmiAddressPools(all: String) =
+    GetPoolsAction(Truthy(all), Permissions.IpAddressApi.UpdateAddress, this)
```
byxorna: this permission seems incorrect. We should use a separate "ipmi.get address pools" perm, and perhaps a separate one for ipmi network view from normal address pool view?

app/collins/controllers/actions/ipaddress/GetPoolsAction.scala
```diff
@@ -38,14 +39,25 @@ case class GetPoolsAction(
           } else {
             IpAddresses.getPoolsInUse()
           }
-        format(pools)
+        val ipmiPools: Set[String] =
+          if (all) {
+            IpmiInfo.AddressConfig.map(_.poolNames).getOrElse(Set())
```
byxorna: Check that this handles mapping the default pool to the proper name correctly? I.e. if you have a config that uses an implicit network (not using the pools = {...} syntax)

Also, does this stuff need to be reverted now that you have a separate endpoint for ipmi pools?

app/collins/controllers/actions/ipmi/GetPoolsAction.scala
```diff
@@ -0,0 +1,86 @@
...
+  case class ActionDataHolder(all: Boolean) extends RequestDataHolder
+
+  override def validate(): Validation = Right(ActionDataHolder(allPools.toBoolean))
+
+  override def execute(rd: RequestDataHolder) = Future {
+    rd match {
+      case ActionDataHolder(all) =>
+        val pools: Set[String] =
+          if (all) {
+            IpmiInfo.AddressConfig.map(_.poolNames).getOrElse(Set())
```
byxorna: both arms of the conditional seem the same. Whats the purpose of the all param then?

app/collins/controllers/actions/ipmi/GetPoolsAction.scala
```diff
@@ -0,0 +1,86 @@
...
+        "SPECIFIED_GATEWAY" -> JsString(pool.gateway.getOrElse("Unspecified")),
+        "GATEWAY" -> JsString(pool.gateway.getOrElse(pool.ipCalc.minAddress)),
+        "BROADCAST" -> JsString(pool.ipCalc.broadcastAddress),
+        "POSSIBLE_ADDRESSES" -> JsNumber(pool.ipCalc.addressCount)
+      )
+      JsObject(seq)
+    }
+  )
+
+  protected def formatNetworkAddress(network: String): String = network match {
```
byxorna: would be cool to refactor this to return an Option instead of a string error message, and do the user-view formatting at a higher level with `flatMap` or something to handle the nones

app/collins/controllers/actions/ipmi/GetPoolsAction.scala
```diff
@@ -0,0 +1,114 @@
...
+case class GetPoolsAction(
+  spec: SecuritySpecification,
+  handler: SecureController
+) extends SecureAction(spec, handler) with AddressActionHelper {
+
+  case class ActionDataHolder(pools: Set[String]) extends RequestDataHolder
+
+  override def validate(): Validation = {
+    var pools: Set[String] = {
+      val p = IpmiInfo.AddressConfig.filter(_.poolNames != Set())
```
byxorna: could probably be `_.poolNames.isNonEmpty` or something

app/collins/controllers/actions/ipmi/GetPoolsAction.scala
```diff
@@ -0,0 +1,114 @@
...
+      if (p.isDefined) {
```
byxorna: i think this could be more idiomatic like 
```
IpmiInfo.AddressConfig.filter(_.poolNames.isNonEmpty) match {
case Set() => Set(IpmiInfo.AddressConfig.flatMap(_.defaultPoolName).getOrElse(""))
case _ => IpmiInfo.AddressConfig.map(_.poolNames).getOrElse(Set(""))
}
``` (syntax may be screwed up, but you get the drift)

Just a suggestion, feel free to disregard

File: app/collins/controllers/actions/ipmi/GetPoolsAction.scala [hunk 8]
```diff
@@ -0,0 +1,114 @@
...
+        IpmiInfo.AddressConfig.map(_.poolNames).getOrElse(Set(""))
+      } else {
+        Set(IpmiInfo.AddressConfig.flatMap(_.defaultPoolName).getOrElse(""))
```
byxorna: So, if you do not specify any pools to pull back, it only returns the default pool? What is the deal iwht the `getOrElse`?

test/collins/controllers/IpmiApiSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+import org.specs2.matcher.JsonMatchers
+
+class IpmiApiSpec extends mutable.Specification with ControllerSpec {
+
+  "IPMI API Specification".title
+
+  args(sequential = true)
+
+  "the REST API" should {
+    "Support getting ipmi pools" in new WithApplication with AssetApiHelper {
```
byxorna: should probably test with and without the optional parameters?

app/collins/controllers/actions/ipmi/GetPoolsAction.scala
michaeljs1990: This is much nicer. I thought i tried this originally but I must have had some other error popping up the first time :(

app/collins/controllers/actions/ipmi/GetPoolsAction.scala
michaeljs1990: I am a pattern matching noob but this does make things look way nicer.

app/collins/controllers/actions/ipmi/GetPoolsAction.scala
michaeljs1990: Yeah if no pools are defined in the configuration it trys to pull back the DEFAULT pool. The get or else was just because this returns an option and I felt dirty just using get.

test/collins/controllers/IpmiApiSpec.scala
michaeljs1990: yes that would be best, I'll see if i can figure out how to switch out or change the configuration file it's reading from so both cases can be tested.

byxorna: The `WithApplication` can use something like `FakeApplication` and provide individual parameters? However, ive not had great experience overriding a config this way (i gave up overriding the IPAM pool configs :()

byxorna:
@michaeljs1990 this looks pretty awesome. It needs a) doc updates, and b) updates to the ruby client and collins-shell to make this accessible.

michaeljs1990:
Added collins shell support in the form of...

```
$ ./bin/collins-shell ipmi pools
Pool       Network         Gateway      Broadcast      Possible Addresses  Start Address  Specified Gateway
OOB-POD01  172.16.32.0/20  172.16.32.1  172.16.47.255                4094  172.16.32.20   Unspecified
```

michaeljs1990:
didn't bump the version since that would cause a conflict with the diff you are about to land.

michaeljs1990:
Thanks for suffering through me learning some Scala 😆 The last rebase i just did just fixes a lint error when i had a backet off by one space in `test/collins/controllers/IpamApiSpec.scala`

byxorna:
@michaeljs1990 looks like my changes i landed created merge conflicts :( soz!

michaeljs1990:
Thats ok :) was expecting it. Going to test locally github gave me a cool GUI conflict resolver that I had to try out.

I believe after that is done though someone else will have to land this as even before the conflict the only option I had was to close the PR.

michaeljs1990:
hmmm, something broke with how the pool name is determined. Looking into it right now... yeah for unit tests.

michaeljs1990:
Ok, this should be all set for landing 🔥 🔥

byxorna:
@michaeljs1990 dope, thanks! Landing now. cc @roymarantz @defect

michaeljs1990:
awww yis! thanks for landing!

PR #523, opened 2017-03-09
make vlan-id attribute optional in lldp

### LINKED ISSUE #466: Collins should handle missing vlan-name in LLDP reports

byxorna:
Fixes #466

This codifies the behavior of some odd switches (some variants of Arista) where the `vlan-id` attribute does not get set in `lldpctl` output, despite the interface definitely being configured with a VLAN. Support this case and don't reject an lldp report just because the asset is in a topology collins doesnt like. 

@michaeljs1990 this is an upstream fix for the hacks we have in Genesis to post-process the LLDP report to fix malformed reports at induction time.

@roymarantz @michaeljs1990 @qx-xp @gtorre @defect

### DIFF
--- a/app/collins/util/config/LldpConfig.scala
+++ b/app/collins/util/config/LldpConfig.scala
@@ -5,8 +5,10 @@ object LldpConfig extends Configurable {
   override val referenceConfigFilename = "lldp_reference.conf"
 
   def requireVlanName = getBoolean("requireVlanName", true)
+  def requireVlanId = getBoolean("requireVlanId", false)
 
   override protected def validateConfig() {
     requireVlanName
+    requireVlanId
   }
 }
--- a/app/collins/util/parsers/LldpParser.scala
+++ b/app/collins/util/parsers/LldpParser.scala
@@ -68,13 +68,13 @@ class LldpParser(txt: String) extends CommonParser[LldpRepresentation](txt) {
 
   protected def findVlans(seq: NodeSeq): Seq[Vlan] = {
     (seq \\ "vlan").foldLeft(Seq[Vlan]()) {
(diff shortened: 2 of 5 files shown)


defect:
Those aristas man... Looks good, but what do you think of adding this as a config parameter? The way we do with `requireVlanName` right now: https://github.com/tumblr/collins/blob/master/conf/reference/lldp_reference.conf#L4

byxorna:
@defect yea, I can totally do that. Ill update with that fix in a few

byxorna:
Tests pass, except for the lame collins-state gem which i need to fix. Landing this and accompanying #530 docs

Issue #528, opened 2017-03-10
[solr] Indexing asset log fails if asset has not yet been indexed


byxorna:
If you create a new asset, and then immediately generate some logs on the asset before the asset doc has been fully indexed by Solr, the `AssetLogSolrUpdater` will raise the following exception, because the AssetLogSerializer is reading `assetTag`, which is defined thusly:

```
   88   @Transient
   89   lazy val assetTag: String = asset.tag
+  90   // TODO(gabe): if this log is indexed (serialized first) before solr is
+  91   // updated with the asset document, this will throw! This can happen when
+  92   // creating a new asset then immediately performing some attribute sets
+  93   // which create logs.
   94   @Transient
   95   lazy val asset: Asset = Asset.findById(assetId).get
```

```
2017-03-10 16:09:36,108 - [ERROR] - SolrPlugin - p.a.LoggerLike$class:error:131 - Added 1 asset documents to be indexed within 200 ms
[ERROR] [03/10/2017 16:09:57.680] [application-akka.actor.default-dispatcher-2] [akka://application/user/change_queue_processor] None.get
java.util.NoSuchElementException: None.get
        at scala.None$.get(Option.scala:347)
        at scala.None$.get(Option.scala:345)
        at collins.models.AssetMetaValue.asset$lzycompute(AssetMetaValue.scala:30)
        at collins.models.AssetMetaValue.asset(AssetMetaValue.scala:30)
        at collins.solr.SolrAssetCallbackHandler.processValue(SolrCallbackHandler.scala:33)
        at collins.solr.SolrAssetCallbackHandler.apply(SolrCallbackHandler.scala:24)
        at collins.callbacks.CallbackManager$$anon$1.propertyChange(CallbackManager.scala:21)
        at java.beans.PropertyChangeSupport.fire(PropertyChangeSupport.java:335)
        at java.beans.PropertyChangeSupport.firePropertyChange(PropertyChangeSupport.java:328)
        at java.beans.PropertyChangeSupport.firePropertyChange(PropertyChangeSupport.java:263)
        at collins.callbacks.CallbackMessageQueue$$anonfun$receive$1.applyOrElse(CallbackMessageQueue.scala:14)
        at akka.actor.Actor$class.aroundReceive(Actor.scala:465)
        at collins.callbacks.CallbackMessageQueue.aroundReceive(CallbackMessageQueue.scala:9)
        at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516)
[...]

PR #529, opened 2017-03-10
Tuning for Solr to improve indexing latency

### LINKED ISSUE #526: Tune solr for realtime searching
Solr is pretty much the default configuration, which causes a bunch of issues at scale, like indexing falling behind and not reflecting changes to assets in searches.

1) Solr indexers fall behind on doc update storms (i.e. set attribute on every asset)
2) commits must be written to disk before being searchable, increasing lag between write and read visibility
3) asset updates may be triggering a manual commit instead of leveraging autocommit to properly batch commits

We should tune for improved autocommit, and look at softautocommit as well. Perhaps using the `commitWithin` solr facility may also provide some relief.

https://wiki.apache.org/solr/SolrConfigXml?#Update_Handler_Section
https://wiki.apache.org/solr/NearRealtimeSearch

byxorna:
Fixes #526 

This PR has a bunch of fixes for how solr behaves to improve indexing latency for assets after write.

1) Removes using explicit `commit()` calls to the Solr server, which reduces the load on solr, given the default 10ms window of batching asset updates (only a handful of assets are reindexed at a time at this interval during mass tag updates).
2) Increased `assetBatchUpdateWindowMs` from 10ms to 30ms to catch more assets in a single batch to add to solr. 
3) Added a `commitWithinMs` tunable (default 50ms) which lets solr decide on the best time to commit updates to the index, instead of us forcing commits (of possibly very small batches of docs). This means that an asset will be available for search in `assetBatchUpdateWindowMs+commitWithinMs=80ms default` from update time.
4) Moves away from hard commits in solr, to using softAutoCommit. This means a document is searchable before solr performs an fsync on the index, reducing the IO load on solr and improving latency between doc update and searchability.

These tunables can be twiddled to balance throughput vs search after write latency for new attributes (i.e. `collins.set_attribute!(tag, :foo, :bar)`, `collins.find(foo: :bar)`). Using the softAutoCommit feature will improve latency (but not entirely remove it!) when updating assets, then immediately searching for them.

@evanelias @bobpattersonjr @defect @roymarantz would love to hear any suggestions! This isnt a perfect fix for the jetpants issue, but its a start.


byxorna:
@defect @roymarantz RFR again?

app/collins/models/AssetLog.scala
```diff
@@ -87,6 +87,10 @@ case class AssetLog(
 
   @Transient
   lazy val assetTag: String = asset.tag
+  // TODO(gabe): if this log is indexed (serialized first) before solr is
+  // updated with the asset document, this will throw! This can happen when
+  // creating a new asset then immediately performing some attribute sets
+  // which create logs.
```
roymarantz: Is there an issue/ticket to fix this?

app/collins/solr/SolrConfig.scala
```diff
@@ -32,8 +32,9 @@ object SolrConfig extends Configurable {
   def socketTimeout = getInt("socketTimeout", 1000)
   def connectionTimeout = getInt("connectionTimeout", 5000)
   def maxTotalConnections = getInt("maxTotalConnections", 100)
+  def commitWithin = getInt("commitWithinMs", 50)
   def defaultMaxConnectionsPerHost = getInt("defaultMaxConnectionsPerHost", 100)
-  def assetBatchUpdateWindow = getInt("assetBatchUpdateWindowMs", 10) milliseconds
+  def assetBatchUpdateWindow = getInt("assetBatchUpdateWindowMs", 30) milliseconds
```
roymarantz: where did these magic numbers come from?

app/collins/solr/SolrConfig.scala
```diff
@@ -32,8 +32,9 @@ object SolrConfig extends Configurable {
   def socketTimeout = getInt("socketTimeout", 1000)
   def connectionTimeout = getInt("connectionTimeout", 5000)
   def maxTotalConnections = getInt("maxTotalConnections", 100)
+  def commitWithin = getInt("commitWithinMs", 50)
```
roymarantz: Are there any docs explaining what commitWithinMs is about?

app/collins/solr/SolrHelper.scala
```diff
@@ -111,35 +111,30 @@ object SolrHelper {
...
-          } else {
-            logger.info("Indexed %d %ss".format(docs.size, serializer.docType.name.toLowerCase))
-          }
-        }
+        server.add(fuckingJava, SolrConfig.commitWithin)
+        logger.debug("Added %d %s documents to be indexed within %d ms".format(
+          fuckingJava.size,
+          serializer.docType.name.toLowerCase,
+          SolrConfig.commitWithin))
+        // dont explicitly hard commit, let solr figure it out and make docs available
```
roymarantz: are there any downsides to doing this?

app/collins/solr/SolrUpdater.scala
```diff
@@ -66,8 +66,30 @@ class AssetSolrUpdater extends Actor {
...
 class AssetLogSolrUpdater extends Actor {
 
+  private[this] def newLogSet = Collections.newSetFromMap[AssetLog](
+    new ConcurrentHashMap[AssetLog, java.lang.Boolean]())
+  private[this] val logsRef = new AtomicReference(newLogSet)
+  private[this] val logger = Logger("AssetLogSolrUpdater")
+  val scheduled = new AtomicBoolean(false)
+  case object Reindex
+
+  // TODO(gabe): perform batch updating on asset logs as well to prevent as much churn in solr
```
roymarantz: is "asset logs" the same as log: AssetLog
If so is this todo note needed?

app/collins/models/AssetLog.scala
byxorna: https://github.com/tumblr/collins/issues/528

app/collins/solr/SolrConfig.scala
byxorna: same place the first numbers came from (the ether). its discussed in the description of the PR why these numbers were chosen as a default

app/collins/solr/SolrConfig.scala
byxorna: yep, its discussed in bullet 3 in the description of the PR, but also https://lucene.apache.org/solr/5_2_0/solr-solrj/org/apache/solr/client/solrj/SolrClient.html#add(java.util.Collection, int)

app/collins/solr/SolrHelper.scala
byxorna: no, because a hard commit isnt actually synchronous because this is happening via actor in background thread

app/collins/solr/SolrUpdater.scala
byxorna: I think this is unneeded now

roymarantz:
👍

byxorna:
@michaeljs1990 wanna give this a once-over before i land?

byxorna:
also @defect, :+1: ?

michaeljs1990:
:+1: on this. Curious about the jetpants issue mentioned in the summary though but didn't see any open issues about it on that repo.

byxorna:
@michaeljs1990 if you set `FOO=xyz` on an asset, then immediately query for `FOO=xyz`, the solr index update may not yet be finalized (used for queries) from the write, so the query will return an "older" set of assets. A workaround is to either slow down how quickly jetpants writes/queries collins, or add some logic to set unique marker tags (or perhaps use modification time) when updating assets and retry queries for those marker tags until a non-empty set is returned.

@evanelias @roymarantz @defect @bobpattersonjr ill land this in a day unless i hear any vocal objections?

grahamc:
I just updated the container used in some jetpants integration testing, and wanted to report this made a direct improvement.

To improve read after write consistency we've added a read-loop after each write, busy-waiting until we can read bagck what we just wrote. Even with this fix in place, sometimes we would update the secondary role on an asset from A to B and then search for assets where  the secondary role is A, and the asset would be returned. After updating the container, under these circumstances with the RaW busy loop, the asset is no longer returned.

Thank you!

byxorna:
@grahamc glad to hear it! This should be a good first (faux) step towards a truly consistent RW interface :)

app/collins/models/AssetLog.scala
komapa: cc: @evanelias

komapa:
Oh, I cc'ed @evanelias but it seems like @grahamc is already basking in the benefits of this :)

app/collins/models/AssetLog.scala
evanelias: So this could break things if we create a config asset and then immediately perform sets on it? I think jetpants may do that in some cases but can't recall by memory. What exactly would break / what are the implications / is this a new problem?

byxorna: @evanelias this isn't new; it's been lurking since inception. Implications are that the tag set will fall (can't find asset to modify in solr index) until solr populates. Not cataclysmic, but definitely not a great consistency model to expose

Issue #534, opened 2017-03-23
NICs in wrong order unless memory banks meet specific criteria


brandt:
Here's a weird bug.

When processing uploaded `lshw` data, Collins gets the NICs backwards unless there are more than 5 memory banks, of which the last three must be occupied.

## Reproducing

I've included some usable example `lshw` files to reproduce the behavior.  (They're missing a lot of details, but Collins will accept them.)

I used the tumblr/collins:latest Docker image.

To verify order, go to the `#hardware-details` tab of the asset page.  The correct order should be:

1. 52:54:00:00:00:01
2. 52:54:00:00:00:02

### Example Failure

If a host (VM for example) has only one memory bank, its NICs will be listed in reverse order in Collins:

```xml
<!-- NICs in reverse order -->
<node class="system">
  <node id="core">
    <node class="processor" handle="DMI:1400" id="cpu">
      <product>Foo</product>
    </node>
    <node class="bridge">
      <node class="network" id="network:0">
        <serial>52:54:00:00:00:01</serial>
        <capacity>1</capacity>
      </node>
      <node class="network" id="network:1">
        <serial>52:54:00:00:00:02</serial>
        <capacity>1</capacity>
      </node>
    </node>
    <node class="memory" id="memory">
      <node class="memory" id="bank">
        <size>1</size>
      </node>
    </node>
  </node>
</node>
```

### Example Success

However, if there are at least 5 memory banks (of which the last 3 *must be occupied*), it will be in the correct order:

```xml
<!-- NICs in correct order -->
<node class="system">
  <node id="core">
    <node class="processor" handle="DMI:1400" id="cpu">
      <product>Foo</product>
    </node>
    <node class="bridge">
      <node class="network" id="network:0">
        <serial>52:54:00:00:00:01</serial>
        <capacity>1</capacity>
      </node>
      <node class="network" id="network:1">
        <serial>52:54:00:00:00:02</serial>
        <capacity>1</capacity>
      </node>
    </node>
    <node class="memory" id="memory">
      <node class="memory" id="bank:0">
        <size>0</size>
      </node>
      <node class="memory" id="bank:1">
        <size>0</size>
      </node>
      <node class="memory" id="bank:2">
        <size>1</size>
      </node>
      <node class="memory" id="bank:3">
        <size>1</size>
      </node>
      <node class="memory" id="bank:4">
        <size>1</size>
      </node>
    </node>
  </node>
</node>
```

[...]

PR #535, opened 2017-03-26
Order metadata by groupId

michaeljs1990:
This orders metadata by the groupId of that metadata before passing it
onto other functions for processing. Since a Map was previously used you
could end up with your data being rendered in any which order. Now data
provided you map it correctly in the following functions will be displayed
in the proper order.

Report: https://github.com/tumblr/collins/issues/534

### DIFF
--- a/app/collins/models/LshwHelper.scala
+++ b/app/collins/models/LshwHelper.scala
@@ -12,6 +12,8 @@ import collins.util.BitStorageUnit
 import collins.util.ByteStorageUnit
 import collins.util.LshwRepresentation
 
+import scala.collection.immutable.SortedMap 
+
 object LshwHelper extends CommonHelper[LshwRepresentation] {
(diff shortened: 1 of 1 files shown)


michaeljs1990:
All tests still pass but someone else should look at this as i'm not sure if it could have other subtle effects.

byxorna:
This LGTM, but I would definitely like unit tests to ensure expected behavior. Can you contrive a fixture lshw/lldp that illustrates the behavior?

byxorna:
@michaeljs1990 bump, checking on unit tests. Is this merge blocking anything internal for you?

michaeljs1990:
No this one is not. I was actually just bored that night so I tried to fix the issue in #534 since it sounded interesting. I'll likely get some unit tests added for this over the weekend.

PR #537, opened 2017-04-17
Add gpu support

jyundt:
This pull request adds support for capturing GPU information in collins. Currently, it is hardcoded for NVIDIA GPUs.

I wasn't clear if the definitions in AsssetMeta were still in use, so I added `GPU_COUNT` and `GPU_DESCRIPTION` to both AssetMeta.scala and the asset_meta table.


michaeljs1990:
Since we couldn't find a great way to separate the onboard graphics card from the external GPUs we do have it hard coded currently. I am not sure how this should be handled since if we just allow it to parse all the GPUs every box would show up with a count of one and then you get into the case of checking NUM_NVIDIA_GPUS and NUM_AMD_GPUS and whatever other flavor you have.

The whitelist was an easy fix for us but possibly not a great way to handle it for a wider audience.

cc @byxorna @roymarantz @defect

byxorna:
@jyundt @michaeljs1990 a common pattern is to make a configuration array of strings to match against LSHW output to identify external graphics adapters (by product, vendor, etc). Up to you if you wanna make it opt-in or opt-out. I would love this to be configurable via production.conf, with a sane default like matching on Manufacturer field with "AMD","NVIDIA"

conf/evolutions/collins/2.sql
```diff
@@ -58,6 +58,9 @@ INSERT INTO asset_meta VALUES (31, 'LLDP_VLAN_NAME', -1, 'LLDP VLANE Name', 'VLA
 INSERT INTO asset_meta VALUES (32, 'INTERFACE_NAME', -1, 'Interface Name', 'Name of physical interface, e.g. eth0');
 INSERT INTO asset_meta VALUES (33, 'INTERFACE_ADDRESS', 0, 'IP Address', 'Address on interface, e.g. 10.0.0.1');
 
+INSERT INTO asset_meta VALUES (34, 'GPU_COUNT', -1, 'GPU Count', 'Number of physical GPUs in asset');
```
byxorna: so, this should actually be a *new* migration; once an evolution is committed, we never update it. You should be able to upgrade your DB schema by only running the newest evolutions. Please move these into a `$(n+1).sql`, thanks

app/collins/models/AssetMeta.scala
```diff
@@ -198,6 +198,10 @@ object AssetMeta extends Schema with AnormAdapter[AssetMeta] with AssetMetaKeys
     val NicName = Value(32, "INTERFACE_NAME")
     // DO NOT USE - Deprecated
     val NicAddress = Value(33, "INTERFACE_ADDRESS")
+
+    //I know this says Deprecated, but I'm unsure of where to define Gpu vars
```
byxorna: did this not compile without this? I would like to not add magic incantations places if it isnt strictly necessary :)

app/collins/models/LshwHelper.scala
```diff
@@ -86,6 +92,41 @@ object LshwHelper extends CommonHelper[LshwRepresentation] {
...
 
+  protected def reconstructGpu(meta: Map[Int, Seq[MetaWrapper]]): FilteredSeq[Gpu] = {
+    val gpuSeq = meta.foldLeft(Seq[Gpu]()) { case (seq, map) =>
+      val groupId = map._1
+      val wrapSeq = map._2
+      val descr = finder(wrapSeq, GpuDescription, _.toString, "")
+      if (descr.isEmpty) {
+        seq
+      } else {
+        Gpu(descr, "", "") +: seq
```
byxorna: Why are these last 2 params ""?

app/collins/models/LshwHelper.scala
```diff
@@ -86,6 +92,41 @@ object LshwHelper extends CommonHelper[LshwRepresentation] {
     )
   }
 
+  protected def reconstructGpu(meta: Map[Int, Seq[MetaWrapper]]): FilteredSeq[Gpu] = {
```
byxorna: i assume this is copy+paste from another reconstruct function? its really hard to read :/

app/collins/util/parsers/LshwParser.scala
```diff
@@ -79,6 +82,13 @@ class LshwParser(txt: String) extends CommonParser[LshwRepresentation](txt) {
       Cpu(cores, threads, speed, asset.description, asset.product, asset.vendor)
   }
 
+  val gpuMatcher: PartialFunction[NodeSeq, Gpu] = {
+    case n if ((n \ "@class" text) == "display" && (n \\ "vendor" text).contains("NVIDIA Corporation")) => {
```
byxorna: yikes, we cant have this. We should check a config parameter that has a list of valid GPU vendors instead

conf/evolutions/collins/7.sql
```diff
@@ -2,9 +2,9 @@
 
 # --- !Ups
 
-UPDATE asset_meta SET priority = -1 WHERE id=33;
+UPDATE asset_meta SET priority = -1 WHERE id=35;
```
byxorna: :-1: 
dont modify existing evolutions

test/collins/util/parsers/LshwParserSpec.scala
```diff
@@ -492,5 +492,18 @@ class LshwParserSpec extends mutable.Specification {
       }
     } // wonky opterons
 
+    "Parse GPU server" in {
```
byxorna: we should have many new GPU-enabled LSHW fixtures to test this functionality. Can you commit a bunch of sample LSHWs from different hw to have a more robust test suite?

jyundt: I'm working on this, but it will probably have to be a separate PR.

We (and by "we" I mean @michaeljs1990) found that our lshw XML files broke all of the existing model tests in `LshwHelperSpec`. In order to get the new GPU code to pass the tests we had to "add" GPUs to the existing XML test files.

Unfortunately it looks like there is something unique about our hardware. But with the exception of the model tests, everything else seems to work fine with our lshw XML data.

app/collins/models/LshwHelper.scala
jyundt: Yeah, I modeled `reconstructGpu` and `collectGpus` off of the existing CPU/memory/disk/NIC functions. I wasn't sure how consistent you'd want the new GPU code to be compared to the previous CPU/memory/disk/NIC stuff.

michaeljs1990:
https://github.com/tumblr/collins/blob/master/support/ruby/collins-client/lib/collins/asset.rb#L206 will want to add gpus to this as well for consistency and so it's easy to access from the ruby client. Not sure if other stuff will need to be updated as well.

michaeljs1990:
Not sure what this is about... but I was testing this PR inside the provided dockerfile for this repo and even adding the additional config needed for gpuVendors to the production.conf file it will not intake GPUs but does show the additional field in hardware. 

Likely doing something really dumb but should likely figure out exactly what that dumb thing is :/

jyundt:
> Not sure what this is about... but I was testing this PR inside the provided dockerfile for this repo and even adding the additional config needed for gpuVendors to the production.conf file it will not intake GPUs but does show the additional field in hardware.

I think I found the culprit, the docker specific `validations.conf` file is missing the `GpuConfig`.

I'll validate and update my PR.

michaeljs1990:
FWIW we ended up having to deploy this which just happens to be the current master branch with this deployed on-top of it. It's working well so far with the exception that we had to install the NVIDIA driver for our genesis iso to get access to some extra information about the GPUs in /proc/sys by default the lshw returns NVIDIA Corporation for product and vendor.

michaeljs1990:
Something strange is happening on update. I checked by uploading a valid XML file with GPUs and then changing the name of a few things such as the NIC and GPU product. The NIC is updated fine and has a new name however it seems GPU is never updated. 

The farthest I have been able to trace it back is to LshwHelper.updateAsset however it seems to have the correct XML at this point.

app/collins/models/LshwHelper.scala
byxorna: ideally this would be made more legible to make support burden easier going forward

test/collins/util/parsers/LshwParserSpec.scala
byxorna: Hmm... ok. I would definitely like any new functionality (especially that which has the potential to break existing functionality) to have unit test coverage before landing. Are you cool with that??

byxorna:
@jyundt is your latest update fixing the odd behavior that @michaeljs1990 pointed out? Is this diff ready for review now? I want to make sure people dont spend time reviewing a WIP

Thanks for the diff, this is gonna be awesome!!!

app/collins/models/AssetMeta.scala
```diff
@@ -206,9 +206,15 @@ object AssetMeta extends Schema with AnormAdapter[AssetMeta] with AssetMetaKeys
     val BaseProduct = findOrCreateFromName("BASE_PRODUCT")
     val BaseVendor = findOrCreateFromName("BASE_VENDOR")
     val BaseSerial = findOrCreateFromName("BASE_SERIAL")
+    val GpuCount = findOrCreateFromName("GPU_COUNT")
```
byxorna: Can this be computed on reconstruction, instead of storing this? It would make it easier to compute how many NVIDIA vs AMD GPUs there are (if in a mixed env)

app/collins/models/AssetMeta.scala
```diff
@@ -206,9 +206,15 @@ object AssetMeta extends Schema with AnormAdapter[AssetMeta] with AssetMetaKeys
...
+    val GpuDescription = findOrCreateFromName("GPU_DESCRIPTION")
```
byxorna: I think we should have both vendor and description? I believe pci enumeration has fields for both vendor and description?

app/views/asset/show_overview.scala.html
```diff
@@ -204,6 +204,14 @@
           <td>@{if (aa.lshw.hasHyperthreadingEnabled) "Yes" else "No"}</td>
         </tr>
 
+          <th colspan="3">GPU</th>
```
byxorna: whitespace is off here

app/views/asset/show_hwdetails.scala.html
```diff
@@ -73,6 +73,23 @@
...
+    <thead>
+      <tr>
+        <th>Id</th><th>Description</th>
+      </tr>
+    </thead>
+    <tbody>
+    @aa.lshw.gpus.zipWithIndex.map { case(gpu,id) =>
+      <tr>
+        <th>@id</th>
+        <td>@gpu.description</td>
```
byxorna: add vendor column here, and maybe sort by id, so the GPUs show up in enumeration order?

app/collins/util/config/GpuConfig.scala
```diff
@@ -0,0 +1,12 @@
+package collins.util.config
+
+object GpuConfig extends Configurable {
+  override val namespace = "gpu"
+  override val referenceConfigFilename = "gpu_reference.conf"
+
+  def gpuVendors= getStringSet("gpuVendors")
```
byxorna: I would like the default values to be reflected here, instead of relying on the config to include something. I.e. the default unpopulated config should include nvidia, instead of the user being forced to set this config on every installation

conf/test_base.conf
```diff
@@ -282,6 +282,10 @@ lshw {
   lshw.defaultNicCapacity=10000000000
 }
 
+gpu {
+  gpuVendors = ["NVIDIA Corporation"]
```
byxorna: This is a stutter. `gpu.gpuVendors`? Maybe `gpu.supportedVendorStrings` or something to be more explicit

jyundt: Good call, renamed `gpuVendors` -> `supportedVendorStrings`

app/collins/util/config/GpuConfig.scala
jyundt: Added default value of "NVIDIA Corporation".

app/views/asset/show_hwdetails.scala.html
jyundt: Added Description (internally called "product") and Vendor columns.

app/views/asset/show_overview.scala.html
jyundt: Fixed, good catch.

app/collins/models/AssetMeta.scala
jyundt: I added product and vendor. It's a bit confusing because lshw refers to the "label" or "description" of the PCI device as the "product".

app/collins/models/AssetMeta.scala
jyundt: Good call, I removed the extraneous `GPU_COUNT`

test/collins/util/parsers/LshwParserSpec.scala
jyundt: Sorry, I should have clarified: I added two additional tests for `LshwHelperSpec` and `LshwParserSpec` to test the new GPU functionality. These new tests (and corresponding lshw XML files) work as expected.

However, for some reason the lshw XML output from all of our new GPU servers will fail all of the existing collins tests. I validated this on the current `master` branch and `add_gpu_support` branch. 

I've been poking around with this, but there appears to be something 'funky' with our hardware, not the recent changes (or the tests).

app/collins/models/LshwHelper.scala
jyundt: Will do, I didn't change this with my most recent commit (d019c27fb30c1541dbb31de075a834283154060a) but I'll get this cleaned up.

jyundt:
> @jyundt is your latest update fixing the odd behavior that @michaeljs1990 pointed out? Is this diff ready for review now? I want to make sure people dont spend time reviewing a WIP

Yes, I incorporated the change from @michaeljs1990 with f301fd99f9f703ab51b302ccb44ce79d828403ce.

My most recent commit (d019c27fb30c1541dbb31de075a834283154060a) should incorporate all of the requested changes _except_ cleaning up `reconstructGpu` (it's still messy).

test/collins/util/parsers/LshwParserSpec.scala
jyundt: As a sanity check, I grabbed lshw output from a Dell device with one gpu (not one of our regular GPU servers) and added two new tests (there are four GPU tests total). Everything works as expected: 7212e38df521774b6aa584a96529d5efadf4fec8

I'm reasonably confident that there is something off with our GPU servers / our lshw output.

byxorna: I am not surprised. LSHW is notoriously shitty in that it produces different output between versions of software and hardware platforms. Do you have a stack trace of the failed tests, or somewhere to begin digging in? I would be ok with a manually edited XML blob if it makes the tests pass (i.e. gut any lshw sections that are making the test explode as an MVP)

app/collins/util/LshwRepresentation.scala
```diff
@@ -105,6 +113,7 @@ case class LshwRepresentation(
         (cpuCoreCount == other.cpuCoreCount) &&
         (cpuThreadCount == other.cpuThreadCount) &&
         (cpuSpeed == other.cpuSpeed) &&
+        (gpuCount == other.gpuCount) &&
```
byxorna: i am not sure where this method gets used, but comparing GPU counts (as opposed to deep comparison of desc/vendor/product) could produce odd behavior when you change GPUs and update LSHW, but the same number of GPUs are present. Would collins not update the meta attributes in the DB if the old reconstructed LSHW `equals` the newly submitted one?

I think what you have in this diff is fine, but we should test this behavior (induct with GPU A, change to GPU B, induct, ensure LSHW is updated properly) and open an issue if this is a problem

app/collins/util/config/GpuConfig.scala
```diff
@@ -0,0 +1,12 @@
+package collins.util.config
+
+object GpuConfig extends Configurable {
+  override val namespace = "gpu"
+  override val referenceConfigFilename = "gpu_reference.conf"
+
+  def supportedVendorStrings= getStringSet("supportedVendorStrings", Set("NVIDIA Corporation"))
```
byxorna: :fire:

conf/evolutions/collins/14.sql
```diff
@@ -0,0 +1,11 @@
+# --- Add GPU information to asset_meta
+
+# --- !Ups
+
+INSERT INTO asset_meta (name, priority, label, description) VALUES ('GPU_PRODUCT', -1, 'GPU Product', 'GPU product (description of GPU)');
```
byxorna: doesnt this need a GPU_DESCRIPTION as well? I see we extract description from LSHW, but it doesnt seem like it ever gets persisted

conf/evolutions/collins/7.sql
byxorna: @jyundt can you undo these modifications?

test/collins/util/parsers/LshwParserSpec.scala
byxorna: Looks like the tests are failing because you are asserting a `Vector()==List()`. @jyundt

```
Expected: LshwRepresentation(Vector(Cpu(8,16....
Actual:      LshwRepresentation(List     (Cpu(8,16...
```

I think if you sort that out (and ensure that `equals()` on `Cpu` and `Gpu` classes work with value equality and not object identity), it should work!

conf/evolutions/collins/14.sql
byxorna: Yea, that is fair. If its not a useful field, lets ignore it. I was thinking it could contain some more useful information, but apparently not! :dancer:

test/collins/util/parsers/LshwParserSpec.scala
jyundt: I noticed the `vector` vs `list` discrepancy previously, but I _just_ stumbled onto the problem: our CPU frequency is different per processor unlike the existing XML test cases.

I manually edited `<size units="Hz">###</size>` from one of our sample XML files and all tests are now passing from our funky gpu servers.

Interestingly enough, I think this problem would have bubbled up previously if `lshw-b0216.xml` was used as a test case for the `LshwHelperSpec` test (instead of just `LshwParserSpec`).

Thanks for the help, lemme take another look at this.

jyundt: @byxorna I poked around this a bit more and I think our usage of dynamic frequency scaling is causing this behavior. The reported CPU speed is "changing" in the lshw XML files depending on the load of the system. 

Additionally, it looks like the current default behavior is to capture CPU information from the first socket and persist that information for all sockets.

As a result of these two behaviors, the existing `LshwHelperSpec` tests will always fail.

TBH, I'm not sure what is "expected behavior" for this. Should collins assume the max clock speed for `CpuSpeedGhz` for all sockets? Should collins honor the data in the lshw XML files and permit different clock speeds per socket?

I'm leaning towards opening a separate GH issue for this because it is unrelated to any of the GPU changes.

Thoughts?

byxorna: I think putting more assumptions into collins to address [mis]configuration of servers during lshw upload is a road we do not want to go further down. IMHO, speed stepping should be disabled by genesis (if it isnt already) before lshw collection, as a workaround. This definitely merits more discussion though, because of the assumptions baked into collins. Could you open an issue and tag me in it? Thanks!

byxorna:
I think this is good to land. @jyundt can you open a new diff for gh-pages to document the new configuration bits for the GPU section?

test/collins/util/parsers/LshwParserSpec.scala
jyundt: Good call with disabling speed stepping, I hadn't thought of that.

New GH issue: https://github.com/tumblr/collins/issues/550

jyundt:
> I think this is good to land. @jyundt can you open a new diff for gh-pages to document the new configuration bits for the GPU section?

#551, let me know if you'd like anything documented/worded differently.

byxorna:
@jyundt looks great! Lets wait for @defect and @roymarantz to review before we land?

app/collins/util/config/GpuConfig.scala
defect: What do you think about moving the default string to the reference config?

defect:
One teeny tiny comment, but looks good to me :)

app/collins/util/config/GpuConfig.scala
byxorna: TBH i feel like the reference config is a confusing layer of indirection. When trying to figure out how collins gets some default config setting, you have to unwind the magic config loading behavior (reference configs, included base configs, environment configs, then application.conf) to understand what gets what from where. I personally prefer (and recommended) to have default config directly in the code

byxorna:
@jyundt ill land this unless you have any last change you need to make, cool?

jyundt:
No, no additional changes. Thanks for your help with this!

PR #538, opened 2017-04-18
Add Classification to asset API

michaeljs1990:
This adds support for finding what classification a nodeclass belongs to using the API.
The tests are a little bit... a lot a bit messy however they cover the two cases well I
believe.

Closes #482 
Docs #541

### DIFF
--- a/test/collins/controllers/AssetApiSpec.scala
+++ b/test/collins/controllers/AssetApiSpec.scala
@@ -4,6 +4,7 @@ import org.specs2.mutable
 
 import play.api.libs.json.JsString
 import play.api.libs.json.Json
+import play.api.mvc.AnyContentAsFormUrlEncoded
 import play.api.test.FakeApplication
 import play.api.test.WithApplication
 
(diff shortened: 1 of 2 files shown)


test/collins/controllers/AssetApiSpec.scala
```diff
@@ -45,6 +46,80 @@ class AssetApiSpec extends mutable.Specification with ControllerSpec with Resour
       }
     }
 
+    "Supports Node Classifier in API" in {
+      "create a configuration asset and check that an asset matches" in new WithApplication(FakeApplication(
+        additionalConfiguration = Map(
+          "solr.enabled" -> false))) with AssetApiHelper {
+        override val assetTag = "node_classy"
```
michaeljs1990: I only have nodes with the utmost class.

byxorna: `"classification":"bourgeoise"`

michaeljs1990:
Just a bump on this 😆 we merged the docs but not the code.... that has to be a first for me.

PR #539, opened 2017-04-18
improve power status reporting

gtorre:
@defect @qx-xp @roymarantz rfr

We need to tighten the regex for power status results. We had an incident where ipmitool was returning a string with the word "on" in the result so collins reported servers as on when they were actually unreachable via ipmi at the time.

Regex behaves as expected:

```
scala> val result = """Chassis Power is on
     |
     | SUCCESS"""
result: String =
Chassis Power is on

SUCCESS

scala> result.contains("Chassis Power is on")
res15: Boolean = true

scala> result.contains("Chassis Power is off")
res16: Boolean = false
```

consolr:
```
[gtorre@collins A:UTIL ~]$ sudo /usr/local/bin/consolr2 -t tag --status
Chassis Power is on


SUCCESS
```

ipmitool:
```
-bash-4.1$ ipmitool -H **** -U **** -P **** -I lanplus power status
Chassis Power is on
```

### DIFF
--- a/app/collins/controllers/actions/asset/PowerStatusAction.scala
+++ b/app/collins/controllers/actions/asset/PowerStatusAction.scala
@@ -31,9 +31,12 @@ case class PowerStatusAction(
   }
   override protected def onSuccess(s: CommandResult): ResponseData = {
     logSuccessfulPowerEvent()
-    val status = s.stdout.contains("on") match {
-      case true => "on"
-      case false => "off"
+    val result = s.stdout
+    val status = "error"
+    if (result.contains("Chassis Power is on")) {
+      val status = "on"
+    } else if (result.contains("Chassis Power is off")) {
+      val status = "off"
     }
     ResponseData(Status.Ok, JsObject(Seq("MESSAGE" -> JsString(status))))
   }


app/collins/controllers/actions/asset/PowerStatusAction.scala
```diff
@@ -31,9 +31,12 @@ case class PowerStatusAction(
...
     logSuccessfulPowerEvent()
-    val status = s.stdout.contains("on") match {
-      case true => "on"
-      case false => "off"
+    val result = s.stdout
+    val status = "error"
+    if (result.contains("Chassis Power is on")) {
+      val status = "on"
+    } else if (result.contains("Chassis Power is off")) {
+      val status = "off"
```
michaeljs1990: This can be reduced to

```scala
val status = s.stdout match {
  case x if x. contains("Chassis Power is on") => "on"
  case x if x. contains("Chassis Power is off") => "off"
  case _ => "error"
}
```

possibly farther but my scala is not great.

roymarantz: I agree with michaeljs1990 unless you have a reason to prefer what you wrote.  I find the suggestion to be closer to the original code and clearer since it introduces fewer variables and syntax

PR #541, opened 2017-04-21
Update docs with new classification json example

michaeljs1990:
Tested it out locally made sure everything ran with `bundle exec rake serve`. I looked around to make sure no other docs need updating but this was the only place I found that looks like it would benefit from additional documentation.

### DIFF
--- a/_data/api.yml
+++ b/_data/api.yml
@@ -429,6 +429,16 @@ sections:
                       }
                     ]
                   },
+                  "CLASSIFICATION": {
+                    "ID": 3,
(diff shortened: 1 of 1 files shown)


byxorna:
its odd that all the details on the classifier live in [the configuration section](https://tumblr.github.io/collins/configuration.html#node%20classifier). Where do you think it would be more useful? I feel like there should be some explicit callouts in the docs saying what to expect back from the API when an asset is classified. @michaeljs1990

michaeljs1990:
🎂 🎈

Issue #546, opened 2017-05-16
DynamicEnum values not being cleaned up


michaeljs1990:
At some point in time AssetMeta.Enum was deprecated however the new AssetMeta.DynamicEnum which took it's place was never fully implemented (at least in my opinion). Currently when you update an asset all of the LSHW values and LLDP values in AssetMeta.Enum are remove from the asset and repopulated however if you add something to the new AssetMeta.DynamicEnum the value is added to the database however the initial value is not removed causing it to be returned instead. From the GUI this looks like the LSHW parser has a bug and is not populating the asset. It also has the undesired side effect of a ton of extra entries in the asset_meta_value db. When i found this bug on a node I had been playing with I had about 50 entries for GPUs which would be a super sweet machine but it only had 4 :(

I have a fix for this I'll post shortly and hope to get merged in :)

@byxorna @defect @roymarantz

michaeljs1990:
I should update this slightly... While we ran into this as being an issue when adding the GPU feature it is also an issue for 

```
BASE_DESCRIPTION
BASE_PRODUCT
BASE_VENDOR
BASE_SERIAL
```

Which end up cropping up strangely in the UI with id's next to them. The nice thing about this fix is the first time a node is inducted again all previous values are cleared and only the new one will remain.

PR #547, opened 2017-05-16
Dynamic Enum Fix

michaeljs1990:
Adds support so that when a value that is of type Dynamic Enum is updated
after initial intake the values are cleared from the initial run and the
new ones are added.

See #546


michaeljs1990:
This needs tests still but I wanted to make sure others were ok with this change before doing so.

app/collins/models/AssetMeta.scala
```diff
@@ -201,6 +201,7 @@ object AssetMeta extends Schema with AnormAdapter[AssetMeta] with AssetMetaKeys
   }
 
   // Post enum fields, enum is not safe to extend with new values
+  type DynamicEnum = AssetMeta
```
michaeljs1990: will remove this... please ignore.

byxorna:
@michaeljs1990 what kind of test cases can be added to prove behavior?

michaeljs1990:
I was going to first induct a machine and then update the asset with the same LSHW except for passing in a different value for BaseDescription, BaseProduct, BaseVendor, and BaseSerial in the xml. Currently it would show the original values from the initial induction process as well as having two entries in the database for each of the above attached to the asset.

byxorna:
@michaeljs1990 if we can do the induction in a test, that would be ideal. That way we know nothing breaks now, and in the future with more changes that will be likely made to this codebase.

michaeljs1990:
I think the added tests should cover all cases. Wasn't sure where to put this test though it makes use of the controller helpers a bunch but it's actually testing a sub section of AssetMeta.

test/collins/controllers/AssetApiHelper.scala
```diff
@@ -54,9 +54,9 @@ trait AssetApiHelper extends ResponseMatchHelpers with JsonMatchers with Control
     result
   }
 
-  def updateHwInfo() = {
+  def updateHwInfo(lshwXml: String = "lshw-basic.xml") = {
     // update hw details (lshw / lldp) - cannot proceed without this
-    val lshwData = getResource("lshw-basic.xml")
+    val lshwData = getResource(lshwXml)
     val lldpData = getResource("lldpctl-two-nic.xml")
```
byxorna: can this be parameterized, as well? While you are in there :)

michaeljs1990: ha, for sure. I can get a second chance at travis not snuffing me on the solr test as well :laughing:

byxorna:
Ive kicked the tests again, because i *think* `2017-05-21 07:02:08,078 - [ERROR] - SolrKey - p.a.LoggerLike$class:error:131 - Cannot create sortable multivalued key for %s, forcing non-sortable` is caused by a race.

byxorna:
@defect @roymarantz can you take a look? This is good for landing

michaeljs1990:
Been running this in prod for a few weeks now without issue side by side with the latest GPU changes.

byxorna:
If we dont hear anyone reject this in a few days, I'll land this. This is an awesome fix!

michaeljs1990:
🔥 🌶 🐝

PR #554, opened 2017-06-02
Fix ipmi pool api

michaeljs1990:
I messed this up somehow even though I checked it a ton of times. Not sure
how i missed this but I have since learned pattern matching in scala sets
is not that simple. This correctly matches against Sets and non Sets. I
updated the test to check for the case of two pools being returned since it
additionally checks for the case that one pool can be returned.

I spend far too much time trying to get different configs to work with the tests
I had some luck using FakeApplication and setting some additionalConfig for it.
However it turns out that if you have two tests using fake config and they are
both different only the first test that runs fake config takes effect and it applies
to all the other runs...


byxorna:
pinging @roymarantz @defect to review as well.

PR #565, opened 2017-08-01
Add support for NVMe disks in LSHW

defect:
Turns out that (at least some) NVMe flash storage is listed as `storage` in the LSHW output instead of `memory` like some other flash storage devices. This PR updates the `diskMatcher` to look for flash devices under both names.

@byxorna @qx-xp @roymarantz @michaeljs1990


byxorna:
Plz post screenies of the hardware summary of a lshw report containing both pcie and scsi disks?

defect:
Finally got around to grabbing screenshots :)

Asset overview page: https://www.dropbox.com/s/5ggkw6om63aagv3/Screenshot%202017-08-08%2016.12.50.png?dl=0
Asset HW detail page: https://www.dropbox.com/s/j57is8vfjyzjkax/Screenshot%202017-08-08%2016.12.43.png?dl=0

byxorna:
Would be nice to aggregate the pcie disk capacity on overview page, so people dont have to do subtraction to figure out how much flash nvme there is in the box

defect:
@byxorna Most flash drives (at least the ones we've been using) don't have their storage capacity in the lshw output for some stupid reason. Hence the sketchy `flashSize` parameter: https://github.com/tumblr/collins/blob/master/conf/reference/lshw_reference.conf#L7

I don't think it's worth implementing an aggregate flash capacity when it's going to be wrong in most cases anyway?

byxorna:
The delta between total storage and scsi storage makes it look like we know how large the pcie storage is, but are just too lazy to display it. Id rather display the pcie storage count, and separately address that number being incorrect in some installs, personally

defect:
@byxorna Done

app/views/asset/show_overview.scala.html
```diff
@@ -250,6 +250,13 @@
           <td>SCSI Storage</td>
           <td>@aa.lshw.totalStorage.toHuman</td>
         </tr>
+	@if(aa.lshw.hasFlashStorage != true) {
```
byxorna: is this logic backwards, or am i crazy? shouldnt it be `@if(aa.lshw.hasFlashStorage)`?

defect: Uh, yup. Pushed a commit from testing stuff 🤦‍♂️ One sec, i'll fix it.

Issue #555, opened 2017-06-07
Capture BIOS version information in ServerBase


jyundt:
@byxorna Thoughts on updating the ServerBase model to include a string for BIOS version information? Or would a separate BIOS model be more appropriate?  (E.g. capture additional information like BIOS vendor, BIOS version, BIOS date, etc.)

We are working with our DC teams and realized we need this information in collins. I wanted to get your input before we make any changes / submit a PR.

CC @michaeljs1990 @nshalman

byxorna:
@jyundt I think having more info is certainly useful. I can't think of a reason to not have this information collected.

jyundt:
Ok cool, I'll take a stab at this.

PR #567, opened 2017-09-14
Adding support for tracking Motherboard and BIOS firmware information

jyundt:
I'm open to suggestions on the best way to track this information, but here is an initial stab.

Screenshot of the updated Server Base hardware details view:

image

I added four tests to validate the motherboard/BIOS info is correct. Let me know if there are any other edge cases that would be appropriate.

(For some background our hardware teams are going to start using this information and we'd like to get it into collins.)

Fixes tumblr/collins#555

cc @michaeljs1990 @byxorna

### DIFF
--- a/app/collins/models/lshw/ServerBase.scala
+++ b/app/collins/models/lshw/ServerBase.scala
@@ -9,11 +9,15 @@ import play.api.libs.json.Json
 object ServerBase {
   implicit object ServerbaseFormat extends Format[ServerBase] {
     override def reads(json: JsValue) = JsSuccess(ServerBase(
+      (json \ "MOTHERBOARD").as[String],
+      (json \ "FIRMWARE").as[String],
       (json \ "DESCRIPTION").as[String],
       (json \ "PRODUCT").as[String],
       (json \ "VENDOR").as[String],
       (json \ "SERIAL").asOpt[String]))
     override def writes(serverbase: ServerBase) = JsObject(Seq(
+      "MOTHERBOARD" -> Json.toJson(serverbase.motherboard),
+      "FIRMWARE" -> Json.toJson(serverbase.firmware),
       "DESCRIPTION" -> Json.toJson(serverbase.description),
       "PRODUCT" -> Json.toJson(serverbase.product),
       "VENDOR" -> Json.toJson(serverbase.vendor),
@@ -22,9 +26,9 @@ object ServerBase {
 }
 
 case class ServerBase(
-    description: String = "", product: String = "", vendor: String = "", serial: Option[String] = None) extends LshwAsset {
+    motherboard: String = "", firmware: String = "", description: String = "", product: String = "", vendor: String = "", serial: Option[String] = None) extends LshwAsset {
   import ServerBase._
   override def toJsValue() = Json.toJson(this)
 
-  def isEmpty(): Boolean = description.isEmpty && product.isEmpty && vendor.isEmpty && serial.isEmpty
+  def isEmpty(): Boolean = motherboard.isEmpty && firmware.isEmpty && description.isEmpty && product.isEmpty && vendor.isEmpty && serial.isEmpty
 }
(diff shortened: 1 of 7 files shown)


jyundt:
cc @nshalman

nshalman:
IIRC someone from hardware suggested tracking the firmware build date as well.

jyundt:
> IIRC someone from hardware suggested tracking the firmware build date as well.

Yeah I _just_ realized I didn't include that. I'll update this shortly.

byxorna:
I am not sure where the line lies, but at some point collecting "facts" about your hardware during induction in genesis and setting normal attributes may be less resistance than needing to extract info from LSHW and bake straight into collins' hardware details. For bios/board model/version, its definitely useful to have in the hardware details. For other, more esoteric fields, I'd err more on the you-set-whatever-attributes-you-want-during-induction side :P

app/collins/models/lshw/ServerBase.scala
```diff
@@ -22,9 +26,9 @@ object ServerBase {
 }
 
 case class ServerBase(
-    description: String = "", product: String = "", vendor: String = "", serial: Option[String] = None) extends LshwAsset {
+    motherboard: String = "", firmware: String = "", description: String = "", product: String = "", vendor: String = "", serial: Option[String] = None) extends LshwAsset {
```
byxorna: why not make these options instead of strings with empty defaults?

app/collins/util/parsers/LshwParser.scala
```diff
@@ -197,13 +197,29 @@ class LshwParser(txt: String) extends CommonParser[LshwRepresentation](txt) {
       val asset = getAsset(elem)
       // serial may be missing, so be flexible here and allow it to be absent
       val serial = (elem \ "serial" headOption).map(_.text)
-      ServerBase(asset.description, asset.product, asset.vendor, serial)
+      val motherboard = (elem \ "node" \ "product").headOption match {
```
byxorna: using Option[String] would simplify this boilerplate logic (/see line 199)

app/views/asset/show_hwdetails.scala.html
```diff
@@ -26,10 +26,15 @@
...
         </tr>
-
         <tr>
             <td>Serial</td><td>@aa.lshw.base.serial</td>
         </tr>
+        <tr>
+            <td>Motherboard</td><td>@aa.lshw.base.motherboard</td>
+        </tr>
+        <tr>
+            <td>Firmware</td><td>@aa.lshw.base.firmware</td>
```
byxorna: its not clear if this is firmware for mobo, ipmi module, bios, or some other bit. Can you clarify, maybe in a `title=""`?

jyundt: I agree, it's not very descriptive. 

I might play around with this a bit more:

image

app/collins/util/parsers/LshwParser.scala
jyundt: Sounds good to me, changed in e64829b54711877d6342e15b8e9019a4b0440eb8.

app/collins/models/lshw/ServerBase.scala
jyundt: Sounds good to me, changed in e64829b54711877d6342e15b8e9019a4b0440eb8.

byxorna:
@jyundt i like it. anything else that would be generally useful that we should get into this evolution before landing this? cc @roymarantz @defect @gtorre @qx-xp

conf/evolutions/collins/15.sql
```diff
@@ -0,0 +1,13 @@
+# --- Add GPU information to asset_meta
```
defect: 👀  Mind updating this to match what it does? :)

defect:
Only took a quick glance but i like it! I'll fire it up tomorrow and try it out a little more closely.

conf/evolutions/collins/15.sql
jyundt: :man_facepalming: 

Good catch.

byxorna:
@defect lets land this?

defect:
Yeah, definitely. I wanted to run the LSHW output from a few types of hardware through it to see what happened, and of course forgot about it. I'll do it first thing tomorrow (don't have my work laptop with me) and merge it.

PR #573, opened 2017-10-11
Update various dependencies to slightly later versions

defect:
Mostly I just want to upgrade the mysql-connector versions prior to 5.1.40 are vulnerable to an exploit that was patched in April. While I was at it i bumped a couple of versions (mainly patch versions) while making sure it still builds and tests pass. I'll need to do a bit more smoke testing though.

The only code update i had to do was to update a method call since `getAddressCount` was deprecated in `commons-net` in favor of `getAddressCountLong`.

At some point I'd like to update to a play version newer than 2.3.x, but I believe they dropped java 7 support in 2.4.x.

@tumblr/collins


byxorna:
@defect why is jvm7 a requirement? the docker image uses 8 fine

defect:
@byxorna It's not, just that we're saying we support java 7 in the docs so we'll probably need a major version bump when we drop it :)

PR #578, opened 2018-03-06
Update the thor dependency constraint for collins-shell

matthiasr:
I need to install it alongside test-kitchen, which depends on <~ 0.19.0,
so do the same here.

We have been using this internally for a little while now and it works fine. I have not tested even newer thor versions.

### DIFF
--- a/support/ruby/collins-shell/Gemfile.lock
+++ b/support/ruby/collins-shell/Gemfile.lock
@@ -2,12 +2,12 @@ PATH
   remote: .
   specs:
     collins_shell (0.2.24)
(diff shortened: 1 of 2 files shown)


matthiasr:
The test job that is failing is the one that's not even running Ruby … not sure if I can do anything about it?

defect:
Yeah, sorry about that. The scala tests are a bit flakey. I'll open an issue about moving the ruby gems to their own repo(s).

This LGTM 👍 

EDIT: Reran the non-ruby job and it finished

PR #579, opened 2018-03-06
Fix the ::Fixnum deprecation warning on Ruby 2.4+

matthiasr:
Fixnum has been [replaced](https://www.ruby-lang.org/en/news/2016/12/25/ruby-2-4-0-released/) by Integer (which used to be a parent class), triggering a very noisy warning.

It is actually safe to add the `to_human_size` method to Integer even on older Ruby version, so this change does that. To get confidence in this I needed to add some tests, so I needed to set up testing for collins-shell in the first place, which in turn required some Gem fixes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,8 @@ matrix:
     - jdk: oraclejdk7
       env: RUN_RUBY=true RUN_SCALA=false GEM_NAME=collins-client
     - jdk: oraclejdk7
+      env: RUN_RUBY=true RUN_SCALA=false GEM_NAME=collins-shell
+    - jdk: oraclejdk7
       env: RUN_RUBY=true RUN_SCALA=false GEM_NAME=collins-state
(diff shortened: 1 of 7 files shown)


defect:
Thanks again! I'm going to push a PR that updates the yard dependencies soon. I'll merge this and #578 at the same time, and bump the patch version.

defect:
@matthiasr Looks like there's a conflict between this and #578. Could you rebase this on current master, or grab https://github.com/defect/collins/commit/0acc22b2aa1e0d3a7cf513c0ab04891843684663 which i think resolves it cleanly?

matthiasr:
resolved! I think a general `bundle update` is also in order but I can submit that separately once this is merged (or you do it?)

matthiasr:
(I also enabled "Allow edits from maintainers" so you should be able to push into my branch)

PR #596, opened 2019-02-18
Add support for deleting a single IP

michaeljs1990:
This lets you specify the IP you would like to delete from an asset
without needing to know the pool of that asset. Stange enough this is
already documented in the API however it was not implemented.

http://tumblr.github.io/collins/api.html#api-ipam-update-an-address

### DIFF
--- a/app/collins/controllers/actions/ipaddress/DeleteAction.scala
+++ b/app/collins/controllers/actions/ipaddress/DeleteAction.scala
@@ -43,15 +43,22 @@ case class DeleteAction(
   }
 
   override def execute(rd: RequestDataHolder) = Future {
-    rd match {
-      case ActionDataHolder(asset, pool, address) =>
-        val deleted = if (address.isDefined) {
-          IpAddresses.deleteByAssetAndAddress(asset, address)
-        } else {
-          IpAddresses.deleteByAssetAndPool(asset, pool)
-        }
-        tattler.notice("Deleted %d IP addresses".format(deleted), asset)
-        ResponseData(Status.Ok, JsObject(Seq("DELETED" -> JsNumber(deleted))))
+    val deleted = rd match {
+      case ActionDataHolder(asset, None, None) => 
+        Right(asset, IpAddresses.deleteByAssetAndPool(asset, None))
+      case ActionDataHolder(asset, None, Some(address)) => 
+        Right(asset, IpAddresses.deleteByAssetAndAddress(asset, Some(address)))
+      case ActionDataHolder(asset, Some(pool), None) => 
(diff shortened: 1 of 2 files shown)


app/collins/controllers/actions/ipaddress/DeleteAction.scala
```diff
@@ -24,24 +25,31 @@ case class DeleteAction(
...
     )
-    Right(ActionDataHolder(asset, pool))
+    Right(ActionDataHolder(asset, pool, address))
   }
 
   override def execute(rd: RequestDataHolder) = Future {
     rd match {
-      case ActionDataHolder(asset, pool) =>
-        val deleted = IpAddresses.deleteByAssetAndPool(asset, pool)
+      case ActionDataHolder(asset, pool, address) =>
```
byxorna: Would it be more "scala-y" for a construct like the following?
```
val deleted = rd match {
  case ActionDataHolder(asset, None, Some(address)) => IpAddresses.deleteByAssetAndAddress(asset, address)
  case ActionDataHolder(asset, Some(pool), None) => IpAddresses.deleteByAssetAndPool(asset, pool)
  case _ => // throw an error cause you can only specify a pool, or an address, not neither or both
}
tattler.notice("Deleted %d IP addresses".format(deleted), asset)
ResponseData(Status.Ok, JsObject(Seq("DELETED" -> JsNumber(deleted))))
```

michaeljs1990: Ah! I was trying to figure out how to do this. This looks much better I'll get this cleaned up.

michaeljs1990: This turned into...
```
       case ActionDataHolder(asset, None, None) => // delete all addresses
      case ActionDataHolder(asset, None, Some(address)) => // delete single address
      case ActionDataHolder(asset, Some(pool), None) => // delete address in single pool
      case ActionDataHolder(asset, Some(pool), Some(address)) => // error thrown
```

The first case I handled because currently in the master branch running
```
curl --basic -X DELETE -u "blake:admin:first" -v 'http://localhost:9000/api/asset/tumblrtag300/addresses'
```
Will trigger a delete on all pools for an asset. If others think we should change that behavior I can switch it but just wanted to make sure that the API change was intentional.

byxorna: Great catch. I think we should document this behavior in the API docs (what happens when you specify none+none, pool, address, or both)

michaeljs1990:
Heh, turns out two years ago someone else wanted to do this also #505 was recently closed and adds this.

PR #598, opened 2019-03-05
Docs for the golang collins cli

michaeljs1990:
Hey, I have been messing around with a golang build for collins CLI for a bit now and have been using it for a while without any major issue outside of a few minor bug fixes. Just thought I would add it here as well in case it's useful for others.

### DIFF
--- a/_includes/tools/collins_go_cli.html
+++ b/_includes/tools/collins_go_cli.html
@@ -1,0 +1,50 @@
+<p class="lead">Simple CLI access to Collins in Golang</p>
+<h2>Overview</h2>
+
+<p><code>collins-go-cli</code> allows you to interact with Collins from the command line. This includes changing attributes, provisioning machines, allocating IPs, maintenance-moding machines, etc. Download it from <a href="https://github.com/michaeljs1990/collins-go-cli/releases">collins-go-cli</a> on Github or if you have golang installed you can <code>go get -u cgit.xrt0x.com/xrt0x/collins-go-cli/cmd/collins</code>. The source can be found at <a href="https://github.com/michaeljs1990/collins-go-cli">github.com/michaeljs1990/collins-go-cli</a>.
+
+<h2>Installation</h2>
+<pre>
+$ go get -u cgit.xrt0x.com/xrt0x/collins-go-cli/cmd/collins
+</pre>
+
+<p>or curl any of the prebuilt binaries from the release page selecting for the appropriate OS.</p>
+
+<pre>
+$ curl -L -o collins https://github.com/michaeljs1990/collins-go-cli/releases/download/2.0.0/collins_linux_amd64
+</pre>
+
+<h2>Setup</h2>
+<p>See <a href="#configs">configs</a> to setup configurations.</p>
+
+<h2>Usage</h2>
+<p>Check out <a href="https://github.com/michaeljs1990/collins-go-cli">github.com/michaeljs1990/collins-go-cli</a> for documentation</p>
+<pre>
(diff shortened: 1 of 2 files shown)


byxorna:
@komapa @defect @alex-laties

_includes/tools/collins_go_cli.html
```diff
@@ -0,0 +1,50 @@
...
+
+<pre>
+$ curl -L -o collins https://github.com/michaeljs1990/collins-go-cli/releases/download/2.0.0/collins_linux_amd64
+</pre>
+
+<h2>Setup</h2>
+<p>See <a href="#configs">configs</a> to setup configurations.</p>
+
+<h2>Usage</h2>
+<p>Check out <a href="https://github.com/michaeljs1990/collins-go-cli">github.com/michaeljs1990/collins-go-cli</a> for documentation</p>
```
byxorna: update this with the output from a `tumblr` org package?

_includes/tools/collins_go_cli.html
```diff
@@ -0,0 +1,50 @@
+<p class="lead">Simple CLI access to Collins in Golang</p>
+<h2>Overview</h2>
+
+<p><code>collins-go-cli</code> allows you to interact with Collins from the command line. This includes changing attributes, provisioning machines, allocating IPs, maintenance-moding machines, etc. Download it from <a href="https://github.com/michaeljs1990/collins-go-cli/releases">collins-go-cli</a> on Github or if you have golang installed you can <code>go get -u cgit.xrt0x.com/xrt0x/collins-go-cli/cmd/collins</code>. The source can be found at <a href="https://github.com/michaeljs1990/collins-go-cli">github.com/michaeljs1990/collins-go-cli</a>.
+
+<h2>Installation</h2>
+<pre>
+$ go get -u cgit.xrt0x.com/xrt0x/collins-go-cli/cmd/collins
```
byxorna: porque no `github.com/tumblr/collins-go-cli`?

byxorna (commented):
lgtm generally! thanks for this!

_includes/tools/collins_go_cli.html
michaeljs1990: ah so the package above it literally a cloned version of your collins-cli ruby gem ported to go with some additional features. I can add docs for the tumblr/go-collins library as well though.

alex-laties:
collins-shell needs to die in a fire. It was unsupported as of 4 years ago... so at this point 🔥.

michaeljs1990:
@defect the feature parity for the collins-go-cli is targeted for you to be able to replace the collins-cli gem if you so wish and pretend like nothing ever happened. You should be able to go back through your bash history and run all of the same commands and get the same outputs. In addition to that I added some features of which I documented some example usage [here](https://github.com/michaeljs1990/collins-go-cli/blob/master/collins/docs/query.txt#L20).

From the main readme these are the changes/additions to the client from the ruby one.

1. JSON output format is a valid JSON dump
2. YAML output format is valid YAML in all languages and not limited to ruby
3. You can pipe IPs, tags, hostnames, or anything else to `collins query` to get more info about them
4. Additional fields that are computed client side. See `func fieldToAssetStruct` or the docs for the full list
5. `collins provision` has a wait flag for that pesky rate limiter
6. `collins power` issues commands inside goroutines so bad BMCs don't slow you down
7. If you have been using the collins ruby cli it should function 100% the same for basic workflows
8. `collins query` has negative matching `key:~valueidontwant` that is available for all flags `-n "~devnode"`

PR #605, opened 2020-01-07
Extract and store logical disk name from lshw data

defect:
This makes the LSHW parser and hw models grab the `logicalname` (i.e. `/dev/sda`, `/dev/sdb` and so on) of each disk from the lshw output. 

This is because lshw doesn't always report them in order, and collins doesn't do a very good job at returning them in order. This makes it tricky to automate things without making a lot of assumptions, and we've had some hardware break those assumptions recently :) Having the logical name attached to each disk should help with automating the disk setup.

If lshw doesn't have the `logicalname` key for some reason it'lll just be an empty string.

I'm also updating the URL to the restlet repo which has apparently moved.

@tumblr/collins

### DIFF
--- a/conf/evolutions/collins/16.sql
+++ b/conf/evolutions/collins/16.sql
@@ -1,0 +1,9 @@
+# --- Add logical name of disks to asset meta
+
+# --- !Ups
+
+INSERT INTO asset_meta (name, priority, label, description) VALUES ('DISK_LOGICAL_NAME', -1, 'Disk logical name', 'The logical name of the disk as reported by lshw');
+
+# --- !Downs
+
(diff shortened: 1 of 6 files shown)


conf/evolutions/collins/16.sql
```diff
@@ -0,0 +1,9 @@
+# --- Add logical name of disks to asset meta
+
+# --- !Ups
+
+INSERT INTO asset_meta (name, priority, label, description) VALUES ('DISK_LOGICAL_NAME', -1, 'Disk logical name', 'The logical name of the disk as reported by lshw');
```
defect: Not sure if this is strictly needed since we're using dynamically created asset meta, but the tests freaked out when not adding it explicitly.

